PR #1, opened 2022-05-30
Add code from haskell-template

srid:
Copies `haskell.nix` from https://github.com/srid/haskell-template/pull/22 and exposes it via the flake.


srid:
Apparently a `flake.lock` is not strictly required as is the case here (the only thing the flake exports is a function).

PR #20, opened 2022-08-22
Add HLS `check`

### LINKED ISSUE #18: Add a `check` for `haskell-language-server`
To sanity check that the project is configured correctly for haskell-language-server to work, we run the following in nix shell:

```
nix develop -i -c haskell-language-server
```

... and check that it successfully exits without any error messages.

The goal of this issue is to have `haskell-flake` automatically add a flake `check` that checks the same. API-wise, it could look like this:

```nix
haskellProjects.foo = {
  checks.haskell-language-server = true;  # True by default if buildTools.haskell-language-server is non-null
};
```

srid:
Resolves #18 

@roberth This is the initial implementation; feel free to critique the diff or the design. :-) I'm pretty sure a few things can be improved.

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -73,12 +73,11 @@ in
                 default = hp: { };
                 defaultText = ''Build tools useful for Haskell development are included by default.'';
               };
-              checks = mkOption {
-                type = types.attrsOf (types.nullOr types.bool);
-                description = ''Built-in checks to enable or disable'';
-                default = {
-                  haskell-language-server = null;
-                };
+              enableHLSCheck = mkOption {
+                type = types.nullOr types.bool;
+                description = ''Whether to enable a flake check to verify if HLS works.'';
+                default = null;
+                defaultText = ''A 'null' value means the check is automatically enabled if buildTools has HLS specified.'';
               };
             };
           });
@@ -130,17 +129,17 @@ in
                         buildTools = (oa.buildTools or [ ]) ++ optionals returnShellEnv buildTools;
                       }));
                   };
-                enableCheckHls =
-                  if cfg.checks.haskell-language-server == null
+                enableHLSCheck =
+                  if cfg.enableHLSCheck == null
                   then lib.attrByPath [ "haskell-language-server" ] null buildTools' != null
-                  else cfg.checks.haskell-language-server;
+                  else cfg.enableHLSCheck;
               in
               rec {
                 package = mkProject { };
                 app = { type = "app"; program = pkgs.lib.getExe package; };
                 devShell = mkProject { returnShellEnv = true; withHoogle = true; };
                 checks =
-                  lib.optionalAttrs enableCheckHls {
+                  lib.optionalAttrs enableHLSCheck {
                     "${projectKey}-hls" =
                       runCommandInSimulatedShell
                         devShell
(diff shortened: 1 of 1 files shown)


flake-module.nix
```diff
@@ -81,9 +88,27 @@ in
...
     perSystem = { config, self', inputs', pkgs, ... }:
       let
+        # Like pkgs.runCommand but runs inside nix-shell with a mutable project directory.
+        #
+        # It currenty respects only the nativeBuildInputs (and no shellHook for
+        # instance), which seems sufficient for our purposes. We also set $HOME and
+        # make the project root mutable, because those are expected when running
+        # something in a project shell (it is indeed the case with HLS).
+        runCommandInNixShell = devShell: projectRoot: name: attrs: command:
+          pkgs.runCommand name (attrs // { buildInputs = devShell.nativeBuildInputs; })
```
roberth, line 98:
```suggestion
          pkgs.runCommand name (attrs // { nativeBuildInputs = devShell.nativeBuildInputs; })
```

flake-module.nix
```diff
@@ -81,9 +88,27 @@ in
...
+            ''
+              # Copy project root to a mutable area
+              # We expect "command" to mutate it.
+              export HOME=$TMP
+              cp -R ${projectRoot} $HOME/project
+              chmod -R a+w $HOME/project
+              pushd $HOME/project
+
+              ${command} > $out
```
roberth, line 107: This is leaky. For example, if `command` is `foo; bar`, only `bar` is redirected.

I'd also make the caller responsible for creating the `$out`.

flake-module.nix
```diff
@@ -81,9 +88,27 @@ in
   config = {
     perSystem = { config, self', inputs', pkgs, ... }:
       let
+        # Like pkgs.runCommand but runs inside nix-shell with a mutable project directory.
+        #
+        # It currenty respects only the nativeBuildInputs (and no shellHook for
+        # instance), which seems sufficient for our purposes. We also set $HOME and
+        # make the project root mutable, because those are expected when running
+        # something in a project shell (it is indeed the case with HLS).
+        runCommandInNixShell = devShell: projectRoot: name: attrs: command:
```
roberth, line 97:
```suggestion
        runCommandInSimulatedShell = devShell: projectRoot: name: attrs: command:
```

flake-module.nix
```diff
@@ -104,11 +130,26 @@ in
...
+                  if cfg.checks.haskell-language-server == null
+                  then lib.attrByPath [ "haskell-language-server" ] null buildTools' != null
+                  else cfg.checks.haskell-language-server;
               in
               rec {
                 package = mkProject { };
                 app = { type = "app"; program = pkgs.lib.getExe package; };
                 devShell = mkProject { returnShellEnv = true; withHoogle = true; };
+                checks =
+                  if enableCheckHls then {
```
roberth, line 143:
```suggestion
                  lib.optionalAttrs enableCheckHls {
```

flake-module.nix
```diff
@@ -73,6 +73,13 @@ in
                 default = hp: { };
                 defaultText = ''Build tools useful for Haskell development are included by default.'';
               };
+              checks = mkOption {
+                type = types.attrsOf (types.nullOr types.bool);
```
roberth, line 77: This is a weird type for a `checks` attribute, but after some reading, the real problem is that it's over-engineered.
I'd suggest an option `enableHLSCheck` with `types.bool` and a good description.

PR #24, opened 2022-09-26
Add a check for `hlint`; rename the HLS check option

### LINKED ISSUE #19: Add a check for `hlint`
Like #18 but for `hlint` (enabled by default *if* `.hlint.yaml` exists at project root and `buildTools.hlint` is non-null).

HariAmoor-professional:
Closes #19

- Add `hlintCheck.enable` option (false by default); adds a flake check to run hlint
- Rename `enableHLSCheck` to `hlsCheck.enable` for consistency
- Internal: dev shell check no longer redirect to `$out` (instead, it simply touches it).

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -97,7 +97,7 @@ in
                     };
                     dirs = mkOption {
                       type = types.listOf types.str;
-                      description = "Directories that should be checked with hlint";
+                      description = "Relative path strings from `root` to directories that should be checked with hlint";
                       default = [ "." ];
                     };
                   };
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -174,7 +174,8 @@ in
                       "${projectKey}-hlint" =
                         runCommandInSimulatedShell
                           devShell
-                          cfg.root "${projectKey}-hlint-check"
+                          cfg.root
+                          "${projectKey}-hlint-check"
                           { buildInputs = [ pkgs.coreutils ]; }
                           ''
                             hlint ${lib.concatStringsSep " " cfg.hlintCheck.dirs}
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -178,6 +178,7 @@ in
                           { buildInputs = [ pkgs.coreutils ]; }
                           ''
                             hlint ${lib.concatStringsSep " " cfg.hlintCheck.dirs}
+                            touch $out
                           ''
                       ;
                     }
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -168,16 +168,19 @@ in
               rec {
                 inherit package devShell;
                 app = { type = "app"; program = pkgs.lib.getExe package; };
-                checks =
-                  (lib.optionalAttrs cfg.hlsCheck.enable {
-                    "${projectKey}-hls" = devShellCheck "hls" "haskell-language-server";
-                  }) // (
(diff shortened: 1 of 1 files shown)


flake-module.nix
```diff
@@ -149,7 +169,25 @@ in
...
                           haskell-language-server > $out
                         '';
-                  };
+                  }) // (
+                    lib.optionalAttrs cfg.hlintCheck.enable {
+                      "${projectKey}-hlint" =
+                        runCommandInSimulatedShell
+                          devShell
+                          cfg.root "${projectKey}-hlint-check"
+                          { buildInputs = [ pkgs.coreutils ]; }
```
roberth, line 178: This would be `nativeBuildInputs`, but I don't think it's even necessary.
```suggestion
                          { }
```

flake-module.nix
```diff
@@ -149,7 +169,25 @@ in
                         ''
                           haskell-language-server > $out
                         '';
-                  };
+                  }) // (
+                    lib.optionalAttrs cfg.hlintCheck.enable {
+                      "${projectKey}-hlint" =
+                        runCommandInSimulatedShell
+                          devShell
+                          cfg.root "${projectKey}-hlint-check"
```
roberth, line 177:
```suggestion
                          cfg.root
                          "${projectKey}-hlint-check"
```

flake-module.nix
```diff
@@ -87,6 +87,22 @@ in
...
+                type = types.submodule {
+                  options = {
+                    enable = mkOption {
+                      type = types.bool;
+                      description = "Check whether hlint is enabled";
+                      default = false;  
+                    };
+                    dirs = mkOption {
+                      type = types.listOf types.str;
+                      description = "Directories that should be checked with hlint";
```
roberth, line 100:
```suggestion
                      description = "Relative path strings from `root` to directories that should be checked with hlint";
```

flake-module.nix
```diff
@@ -87,6 +87,22 @@ in
...
+                      default = [];
```
roberth, line 101:
```suggestion
                      default = ["."];
```

roberth (commented):
Some suggestions, but haven't tried it yet.

flake-module.nix
HariAmoor-professional: A more sensible default would be `src`, which is the root of Srid's [haskell-template](https://github.com/srid/haskell-template) repo; this is the reason that we implemented the whole `find ...` command on line 186.

Do you have a better suggestion? Current directory doesn't make sense b/c you might have random Nix files and stuff that don't apply to `hlint`.

flake-module.nix
HariAmoor-professional: If it's not necessary, then maybe just remove it altogether.

flake-module.nix
srid: I didn't realize that `hlint .` worked. Yea, that would be a better default.

> Current directory doesn't make sense b/c you might have random Nix files and stuff that don't apply to hlint.

hlint will only look at `.hs` files, I assume.

HariAmoor-professional: If that's the case, then I have no issue

srid (approved):
Looks good enough to merge for me. I'm in a hurry. So Robert, if you have any more suggestions, please feel free to leave them (and I can make the change separately).

I'm obliged to rename the HLS check for consistency here.

PR #27, opened 2022-10-16
Multi-package projects

### LINKED ISSUE #7: Multi-package projects
How should this flake be changed to support projects with multiple Haskel packages, like https://github.com/hercules-ci/hercules-ci-agent ?

HariAmoor-professional:
Closes #7

Labeled as draft b/c not yet tested

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -90,7 +90,6 @@ in
                       '';
                       default = false;
                     };
-
                   };
                 };
               };
@@ -214,7 +213,7 @@ in
                     } // (
                       if cfg.topLevelPackages.enableMultiPkgs then {
                         packages = pkgs: map (pkg: pkgs.${pkg}) cfg.topLevelPackages.dirs;
-                        buildInputs = [ hp.ghcid ];
+                        nativeBuildInputs = [ hp.ghcid ];
                       } else { }
                     )
                   );
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -110,6 +110,32 @@ in
                   };
                 };
               };
+              topLevelPackages = mkOption {
+                type = types.submodule {
+                  options = {
+                    enableMultiPkgs = mkBoolOption {
+                      description = "Whether or not to use multiple top-level package dirs";
+                      default = false;
+                    };
+
+                    dirs = mkOption {
+                      type = types.listOf types.str;
+                      description = ''
+                        Subdirs containing a separate Cabal package
+
+                        Define as nonempty only if enableMultiPkgs is set to `true`
+                      '';
+                      default = [ ];
+                    };
+
+                    defaultPkg = mkOption {
+                      type = types.str;
+                      description = "Default package to pass to flake if multiple packages exist";
+                      default = "";
+                    };
+                  };
+                };
+              };
             };
           });
         };
@@ -151,6 +177,23 @@ in
                   (pkgs.lib.composeExtensions
                     (pkgs.haskell.lib.packageSourceOverrides cfg.source-overrides)
                     cfg.overrides);
+                hp = hp.extend
+                  (
+                    let
+                      enableMultiPkgs = cfg.topLevelPackages.enableMultiPkgs;
+                      dirs = cfg.topLevelPackages.dirs;
+                    in
+                    if enableMultiPkgs then
+                      (self: _:
+                        listToAttrs (map
+                          (dir: {
+                            name = dir;
+                            value = self.callCabal2nix dir ./${dir} { };
+                          })
+                          dirs)
+                      )
+                    else (_: _: { })
+                  );
                 defaultBuildTools = hp: with hp; {
                   inherit
                     cabal-install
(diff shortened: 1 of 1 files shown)


flake-module.nix
```diff
@@ -110,6 +110,32 @@ in
                   };
                 };
               };
+              topLevelPackages = mkOption {
+                type = types.submodule {
+                  options = {
+                    enableMultiPkgs = mkBoolOption {
```
roberth, line 116: Could this be inferred from `dirs != []`?

flake-module.nix
```diff
@@ -159,9 +202,22 @@ in
...
                 buildTools = lib.attrValues (defaultBuildTools hp // cfg.buildTools hp);
-                package = cfg.modifier (hp.callCabal2nixWithOptions cfg.name cfg.root "" { });
+                package =
+                  if cfg.topLevelPackages.enableMultiPkgs then
+                    hp.${cfg.topLevelPackages.defaultPkg}
+                  else
+                    cfg.modifier (hp.callCabal2nixWithOptions cfg.name cfg.root "" { });
                 devShell = with pkgs.haskell.lib;
-                  (addBuildTools package buildTools).envFunc { withHoogle = true; };
+                  (addBuildTools package buildTools).envFunc (
```
roberth, line 211: Perhaps instead of `envFunc`, `hp.shellFor` would be nicer, because it merges the environments of multiple packages.

flake-module.nix
```diff
@@ -159,9 +202,22 @@ in
                     hlint;
                 };
                 buildTools = lib.attrValues (defaultBuildTools hp // cfg.buildTools hp);
-                package = cfg.modifier (hp.callCabal2nixWithOptions cfg.name cfg.root "" { });
+                package =
+                  if cfg.topLevelPackages.enableMultiPkgs then
+                    hp.${cfg.topLevelPackages.defaultPkg}
+                  else
+                    cfg.modifier (hp.callCabal2nixWithOptions cfg.name cfg.root "" { });
```
roberth, line 209: I think it's desirable to use the overlay (`haskellPackages.extend`) even in case of a single package, because it removes the possibility of accidentally using the Nixpkgs version of the package instead of the locally defined one.

flake-module.nix
```diff
@@ -110,6 +110,32 @@ in
                   };
                 };
               };
+              topLevelPackages = mkOption {
```
roberth, line 113: I think the `topLevelPackages` namespace shouldn't be necessary.

flake-module.nix
```diff
@@ -159,9 +202,22 @@ in
...
+                    cfg.modifier (hp.callCabal2nixWithOptions cfg.name cfg.root "" { });
                 devShell = with pkgs.haskell.lib;
-                  (addBuildTools package buildTools).envFunc { withHoogle = true; };
+                  (addBuildTools package buildTools).envFunc (
+                    {
+                      withHoogle = true;
+                    } // (
+                      if cfg.topLevelPackages.enableMultiPkgs then {
+                        packages = pkgs: map (pkg: pkgs.${pkg}) cfg.topLevelPackages.dirs;
+                        buildInputs = [ hp.ghcid ];
```
roberth, line 217:
```suggestion
                        nativeBuildInputs = [ hp.ghcid ];
```

This will provide shell completions and cross compilation support.

roberth (commented):
Some thoughts about the design and implementation. I hope it's helpful.

flake-module.nix
HariAmoor-professional: @roberth Then how do you define `defaultPkg`? Or should you not need to?m

roberth: We should not need to define a single package.

Currently `haskellProjects.<proj>` defined a flake `packages` package for each `<proj>`. That 1:1 relation is what `package` is for, but now we'll have multiple packages per project.
For each `haskellProjects.<proj>.dirs.*`, we could have a package named `<proj>-<dirname>`.
Ideally we'd use the actual package name instead of the directory name.

A list of directories is actual a bit "high level". I think a good first step is to create an `attrsOf submodule` for the locally defined packages. For example:
`haskellProjects.<proj>.package.<pkgname>.src`
`haskellProjects.<proj>.package.<pkgname>.packageAttribute` (default: `${proj}-${pkgname}`)

The logic for that should be more straightforward to implement than `dirs`, but you can still write `dirs` to make use of those options and get a good separation of concerns. It also allows a bit more control: customizing the package attribute, and perhaps some extra options to pass to cabal2nix.

srid: Yes, the structure should be more or less like that. I just put a concrete proposal here (should have done it long ago to assist in this PR): https://github.com/srid/haskell-flake/issues/7#issuecomment-1282932587

The `flakeAttribute` part (or `packageAttribute` in the above comment) is still not clear to me. I suppose we can finalize that once the general functionality is in place (in the PR).

HariAmoor-professional:
Closed in favor of a new PR (to be delivered)

PR #34, opened 2022-10-30
Multi-package projects

### LINKED ISSUE #2: Do not create `app` if there is none
Ema is a library:

```
❯ nix run github:srid/ema/multisite
error: unable to execute '/nix/store/44gpjhbn175y00jyy5z1k0jwy6q946wg-ema-0.7.2.0/bin/ema': No such file or directory
```

One solution: we could just add an `executable = true;` option for this.

### LINKED ISSUE #7: Multi-package projects

HariAmoor-professional:
Resolves #7 and resolves #2

~~**Not ready to be merged due to infinite recursion error**~~ Fixed

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -51,12 +51,11 @@ in
               };
             };
           };
-          # Define nested packages
           packageAttrsSubmodule = with types; submodule {
             options = {
               root = mkOption {
                 type = path;
-                description = "Path to Haskell package";
+                description = "Path to Haskell package where the .cabal file lives";
               };
               modifier = mkOption {
                 type = functionTo package;
@@ -91,7 +90,6 @@ in
                 description = ''Overrides for the Cabal project'';
                 default = self: super: { };
               };
-              # TODO: This option will go away after #7
               modifier = mkOption {
                 type = functionTo types.package;
                 description = ''
@@ -121,6 +119,11 @@ in
               };
               packages = mkOption {
                 type = types.lazyAttrsOf packageAttrsSubmodule;
+                description = ''
+                  Attrset of local packages in the project repository.
+
+                  Autodetected by default by looking for `.cabal` files in sub-directories.
+                '';
                 default =
                   lib.mapAttrs
                     (_: value: { root = value; })
@@ -166,12 +169,17 @@ in
             (projectKey: cfg:
               let
                 inherit (pkgs.lib.lists) optionals;
+                projectOverlay =
+                  pkgs.lib.composeExtensions (pkgs.haskell.lib.packageSourceOverrides cfg.source-overrides) cfg.overrides;
                 # Apply user provided source-overrides and overrides to
                 # `cfg.haskellPackages`.
-                hp = cfg.haskellPackages.extend
-                  (pkgs.lib.composeExtensions
-                    (pkgs.haskell.lib.packageSourceOverrides cfg.source-overrides)
-                    cfg.overrides);
+                projectPackages = cfg.haskellPackages.extend projectOverlay;
+                localPackagesOverlay = self: _:
+                  lib.mapAttrs
+                    (name: value: value.modifier (self.callCabal2nix name value.root { }))
+                    cfg.packages;
+                finalPackages = projectPackages.extend localPackagesOverlay;
+
                 defaultBuildTools = hp: with hp; {
                   inherit
                     cabal-install
@@ -179,16 +187,11 @@ in
                     ghcid
                     hlint;
                 };
-                buildTools = lib.attrValues (defaultBuildTools hp // cfg.buildTools hp);
-                nestedPackagesOverlay = self: _:
-                  lib.mapAttrs
-                    (name: value: self.callCabal2nix name value.root { })
-                    cfg.packages;
-                nestedPackagesHp = hp.extend nestedPackagesOverlay;
-                devShell = nestedPackagesHp.shellFor {
+                buildTools = lib.attrValues (defaultBuildTools projectPackages // cfg.buildTools projectPackages);
+                devShell = finalPackages.shellFor {
                   packages = p:
                     map
-                      (name: cfg.packages."${name}".modifier p."${name}")
+                      (name: p."${name}")
                       (lib.attrNames cfg.packages);
                   withHoogle = true;
                   nativeBuildInputs = buildTools;
@@ -200,7 +203,7 @@ in
                 inherit devShell;
                 packages =
                   lib.mapAttrs
-                    (name: value: value.modifier nestedPackagesHp."${name}")
+                    (name: _: finalPackages."${name}")
                     cfg.packages;
                 checks = lib.filterAttrs (_: v: v != null)
                   {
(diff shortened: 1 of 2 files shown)


flake-module.nix
```diff
@@ -113,6 +131,13 @@ in
                 default = { };
                 type = hlintCheckSubmodule;
               };
+              packages = mkOption {
+                type = types.attrsOf packageAttrsSubmodule;
```
roberth, line 135:
```suggestion
                type = types.lazyAttrsOf packageAttrsSubmodule;
```

Typical infinite recursion footgun, if packages depend on packages. I hope this was your problem?

HariAmoor-professional: Unfortunately not; seems there's another bug elsewhere. I'll push my test-case and then we can take a look.

flake-module.nix
```diff
@@ -203,13 +224,25 @@ in
...
+                      name =
+                        # Prefix package names with the project name (unless
+                        # project is named `default`)
+                        if projectName == "default"
+                        then packageName
+                        else "${projectName}-${packageName}";
+                      value = package;
+                    })
+                    project.packages
+                  )
```
srid, line 242: I wonder if the Nix code here can be refactored?

srid (commented):
Functionality-wise, this looks great.

@roberth What do you think about the flake attribute naming? Uses "$project-$package" (unless the project is named "default").

flake-module.nix
```diff
@@ -51,6 +51,19 @@ in
...
+          packageAttrsSubmodule = with types; submodule {
+            options = {
+              root = mkOption {
+                type = path;
+                description = "Path to Haskell package where the .cabal file lives";
+              };
+              modifier = mkOption {
+                type = functionTo package;
+                default = drv: drv;
+                description = "Modifier to be applied to Haskell package before applying to overlay";
```
roberth, line 63:
```suggestion
                description = "Modifier to be applied to Haskell package before adding to overlay";
```

Avoids association with function application.

flake-module.nix
```diff
@@ -203,13 +224,25 @@ in
       in
       {
         packages =
-          lib.mapAttrs
-            (_: project: project.package)
-            projects;
-        apps =
-          lib.mapAttrs
-            (_: project: project.app)
-            projects;
```
roberth, line 212: `apps` were removed. It was not an essential feature afaic, and it was incomplete. It should have filtered for packages with executable, and add an app for each executable. This is non-trivial and can be added later.

roberth:
> @roberth What do you think about the flake attribute naming? Uses "$project-$package" (unless the project is named "default").

lgtm

flake-module.nix
```diff
@@ -153,37 +169,42 @@ in
...
-                hp = cfg.haskellPackages.extend
-                  (pkgs.lib.composeExtensions
-                    (pkgs.haskell.lib.packageSourceOverrides cfg.source-overrides)
-                    cfg.overrides);
+                projectPackages = cfg.haskellPackages.extend projectOverlay;
+                localPackagesOverlay = self: _:
+                  lib.mapAttrs
+                    (name: value: value.modifier (self.callCabal2nix name value.root { }))
+                    cfg.packages;
+                finalPackages = projectPackages.extend localPackagesOverlay;
```
roberth, line 181: I have some package overrides that need to access the package set, so they can't be in `packages.<name>.modifier`, but `overrides` doesn't work either because that is applied before the packages are added.

Perhaps the order of the overlays could be reversed, so that `overrides` applies last?

I recommend to use `composeExtensions` (or `composeManyExtensions`) because multiple uses of `extend` each create a very large attribute set that goes to waste.

srid: > I have some package overrides that need to access the package set, so they can't be in `packages.<name>.modifier`, but `overrides` doesn't work either because that is applied before the packages are added.

Do you have an example use-case?

srid:
I'll improve the README.md for this PR shortly.

flake-module.nix
HariAmoor-professional: Closing in favor of #36

flake-module.nix
HariAmoor-professional: LGTM

flake-module.nix
HariAmoor-professional: Resolving for now, in the absence of something better; @roberth and others can feel free to reopen if there are any suggestions

flake-module.nix
roberth: - `buildFromSdist` (though that would be superseded by #35)
 - optparse applicative completions [since recently](https://github.com/NixOS/nixpkgs/commit/ac1f1ad0e0a8cfd35db476529c82354e033a48cc#diff-f690c88564d6774086ecabe92c7e4c18078889589f8978ad0819f57bf515eaf9)
 - `addBuildToolDepends`
 - `overrideScope` or `.override` to change a dependency for just one package

It's not much of an example, but my real life use case is https://github.com/hercules-ci/hercules-ci-agent/blob/master/nix/packages.nix#L82.

flake-module.nix
roberth: `listToAttrs` after `mapAttrsToList` should be equal to `mapAttrs'`.

The `foldl` can perhaps be changed to `mkMerge` like below.

Perhaps it could be cut up by naming things.

flake-module.nix
srid: I think I understand the problem now.

> Perhaps the order of the overlays could be reversed, so that `overrides` applies last?

Yea, this makes more sense to me, because `overrides` becomes _the_ place where you get "the final say" in overriding things. One could override even the local packages (not just dependencies) in `overrides`. With this change, will there be any value to keeping `packages.${name}.modifier`? Or should we remove the `modifier` option?

roberth: > will there be any value to keeping `packages.${name}.modifier`? Or should we remove the `modifier` option?

Let's remove `modifier`. Having a less powerful way of doing the same thing is just confusing.

It being in `packages.${name}` was nice, but that's not enough reason to keep it.

flake-module.nix
HariAmoor-professional: Done

flake-module.nix
srid: > should I achieve equivalent functionality with overrides?

That's how it should be done, yes.

srid: > Perhaps the order of the overlays could be reversed, so that `overrides` applies last?

Did in 13a2646304e654c603e95620d67ec7f9e9ae8ac7 ... hope this is the correct order:

```nix
                  pkgs.lib.composeManyExtensions 
                    [ # The order here matters.
                      #
                      # User's overrides (cfg.overrides) is applied **last** so
                      # as to give them maximum control over the final package
                      # set used.
                      cfg.overrides
                      (pkgs.haskell.lib.packageSourceOverrides cfg.source-overrides) 
                      localPackagesOverlay
                    ];
```

srid (approved):

Issue #25, opened 2022-09-26
Add automated tests


srid:
What is the best way to add tests for this repo? cc @roberth 

I normally test branches against another branch of https://github.com/srid/haskell-template - but having something automated would be nice. https://github.com/numtide/treefmt-nix/tree/master/tests uses a sub-flake for tests.

roberth:
Doesn't that lock the test's "parent" input?
In flake-parts I went out of my way to avoid depending on the parent (https://github.com/hercules-ci/flake-parts/blob/main/dev/default.nix) but I don't think we should repeat that here. (But maybe I'm wrong about that)
I'm kind of hoping for the [lazy trees](https://github.com/NixOS/nix/pull/6530) to fix the `..` lock problem.

srid:
Pending better ways, I'm inclined to go with the approach of https://github.com/Platonic-Systems/mission-control

We also get an example and flake template for free.

PR #40, opened 2022-11-09
Fix regression with `checks`

srid:
Bug fixes on top of #34

(Yea, we really need #25 )


srid:
Caught in https://github.com/srid/haskell-template/pull/67

Issue #37, opened 2022-11-01
Option to disable the development environment


srid:
Haskell developers will always remain the *primary* audience of `haskell-flake`. For users who just want to package the Haskell app without needing the full-blown development environment, the following configuration is useful:

```nix
haskelProjects.default = {
  buildTools = _: {
    cabal-install = null;
    ghcid = null;
    hlint = null;
    haskell-language-server = null;
  };
}
```

(These just disable the default build tools added by haskell-flake).

However, this configuration does not prevent the addition of `devShell` itself that is going to remain unused by these users.

From an API standpoint, how can we provide a simpler option to disable the whole development shell and friends for this target audience (non-Haskell developer users)?

srid:
Simplest API I can think of,

```nix
haskellProjects.<name>.development = false;
```

... should disable the `devShell` injection entirely, thus all of `buildTools` becomes unused. But since `buildTools` now becomes contingent on this new development flag, perhaps we should move it under that as well, viz.:

```nix
haskellProjects.default = {
  packages = { };
  development = {
    enable = true; # True by default
    buildTools = hp: {
      ghcid = null;
      fourmolu = hp.fourmolu;
    };
  };
```

PR #44, opened 2022-11-10
Add extra shell arguments

roberth:
I think I'll need these, but I haven't tried switching my dev shell yet.

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -96,6 +96,25 @@ in
                 default = hp: { };
                 defaultText = ''Build tools useful for Haskell development are included by default.'';
               };
+              shell.extraHaskellLibraries = mkOption {
+                type = functionTo (types.listOf types.package);
+                description = ''
+                  A list of libraries that are included in the devShell, as if they were dependencies of one of the packages.
+
+                  This is useful for Haskell scripts that rely on the devShell to provide their libraries.
+                '';
+                default = p: [];
+                defaultText = lib.literalExpression "p: [ ]";
+                example = lib.literal
+              };
+              shell.buildInputs = mkOption {
+                type = types.listOf types.package;
+                description = ''
+                  A list of non-Haskell `buildInputs` that are included in the devShell.
+
+                  This is useful for dependencies on C libraries, such as `pkgs.zlib`.
+                '';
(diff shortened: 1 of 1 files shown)


flake-module.nix
```diff
@@ -96,6 +96,25 @@ in
...
+                default = p: [];
+                defaultText = lib.literalExpression "p: [ ]";
+                example = lib.literal
+              };
+              shell.buildInputs = mkOption {
+                type = types.listOf types.package;
+                description = ''
+                  A list of non-Haskell `buildInputs` that are included in the devShell.
+
+                  This is useful for dependencies on C libraries, such as `pkgs.zlib`.
```
srid, line 115: Is there an example Haskell project that does this? I assume that if a Haskell project depends on C library, we would have to add it in *two* places:

1. devShell's buildInputs
2. `addBuildDepends` on the local package itself

If so, wouldn't it be a nicer API to provide a single option that does both?

Emanote does this, but it is a binary dependency (rather than library dependency) for `stork`; (1) is [here](https://github.com/EmaApps/emanote/blob/cc65b533e780f902ca223268c93f5423c489182a/flake.nix#L37-L38) and (2) is [here](https://github.com/EmaApps/emanote/blob/cc65b533e780f902ca223268c93f5423c489182a/flake.nix#L50-L52).

srid:
The `shell.` option prefix is nice, and I think `buildTools` fits inside it as well (`shell.buildTools`)? See also `development.` prefix of #37 (how do we consolidate `shell.` with that?)

flake-module.nix
roberth: hercules-ci-agent does: https://github.com/hercules-ci/hercules-ci-agent/blob/master/nix/shellFor-cabal.nix

Progress converting it so far. No shell yet: https://github.com/hercules-ci/hercules-ci-agent/commit/e023859a16ac1727c55f06f067556fea6b57d5aa

> If so, wouldn't it be a nicer API to provide a single option that does both?

I agree, and this seems like something that shellFor should implement as well.

srid:
@roberth This can now be closed right?

roberth:
That's right!

PR #45, opened 2022-11-11
Option doc improvements

roberth:


roberth:
Deployed: https://flake.parts/options/haskell-flake.html

Issue #21, opened 2022-08-28
HLS check fails by trying to access the network


srid:
The check works on my local machine but not on Github Actions runner or Garnix CI.

https://github.com/srid/haskell-template/runs/8060856223?check_suite_focus=true

```
default-hls-check> Severity: DsError
default-hls-check> Message:
default-hls-check>   Error when calling cabal exec -v0 -- ghc --print-libdir
default-hls-check>   dieVerbatim: user error (cabal: Couldn't establish HTTP connection. Possible cause: HTTP proxy
default-hls-check>   server
default-hls-check>   is down.
default-hls-check>   )
```

Very strange.

srid:
Because this check requires disabling the Nix sandbox for it to work, I think we should disable the `enableHLSCheck` option by default.

srid:
This fails on Hercules CI, understandably: https://hercules-ci.com/github/srid/haskell-template/jobs/100

So, I'd have to `hlsCheck.enable = true;` and add a Hercules CI *effect* that runs the HLS check. For this to be doable, however, `haskell-flake` has to expose the check derivation itself in some form, like `config.haskellProjects.default.hlsCheck.drv`.

PR #51, opened 2022-12-23
Add haskellProjects.*.{hlintCheck,hlsCheck}.drv

srid:
cf. https://github.com/srid/haskell-flake/issues/21#issuecomment-1364154220

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -12,13 +12,36 @@ let
   inherit (types)
     functionTo
     raw;
+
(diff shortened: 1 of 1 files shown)


srid:
This is useless; as we want the whole thing as a string to pass to hci effects.

runCommandInSimulatedShell could perhaps be upstreamed to hci-effects in some form, replacing or complementing its existing `nix-shell` function.

PR #53, opened 2023-01-05
Configure Hercules CI

srid:
I believe `herculesCI.ciSystems` is mandatory to limit the building on these two systems (because that's all I have as agents).

https://docs.hercules-ci.com/hercules-ci-agent/evaluation/#_default_systems

Hercules CI will be used latter for testing #25

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,5 +2,6 @@
   description = "A `flake-parts` module for Haskell development";
   outputs = { self, ... }: {
     flakeModule = ./flake-module.nix;
+    herculesCI.ciSystems = [ "x86_64-linux" "aarch64-darwin" ];
   };
 }


srid:
https://hercules-ci.com/github/srid/haskell-flake/jobs/2

```
Evaluation failed
Please make sure your ci.nix has at least one derivation.
```

Hmm, I don't think this error message is accurate? It assumes I must have a `ci.nix`, but that's not the case. @roberth

srid:
Looks like I need to add some derivation outputs to make HCI happy. Which I have to do anyway, eventually, when adding a `check` for #25 

`flake-parts` uses a [dev flake](https://github.com/hercules-ci/flake-parts/tree/main/dev) approach. Should haskell-flake do the same, or is there a better way?

PR #60, opened 2023-02-01
Add automated test

### LINKED ISSUE #25: Add automated tests

srid:
Resolves #25. To be squash merged.

- Add `./test` that aims to use all features of haskell-flake
- Add `./test/test.sh` that verifies that the devshell of `./test` is as one would expect
- Add `./runtest.sh` at project root to run the full test
- Run the same ^ in Github Actions

I'm *not* entirely happy with this approach; in particular, `test.sh` seems pretty hacky. But it works. Not sure if it will scale.

I just would like to have some sort of automated tests before doing substantial changes to `haskell-flake`.


roberth:
> I'm _not_ entirely happy with this approach

Me neither, but it seems to get the job done.
Improvements are kind of blocked Flakes improvements like
 - https://github.com/NixOS/nix/issues/6124
 - https://github.com/NixOS/nix/issues/7028
 - https://github.com/NixOS/nix/issues/7730
 - possibly https://github.com/NixOS/nix/pull/6530 for `..` input urls?

srid:
Yea, let's go with it for now and make it perfect when the time is right.

Issue #12, opened 2022-06-02
Expose `overrides` in flake


srid:
**Rationale**: When I want to use emanote as a library, I want to use its `overrides` (and `source-overrides`) so as to be able to re-use it without replicating it in the local project (that uses emanote as a dependency).

https://github.com/srid/emanote/blob/6f83b857741fb6607716cf97cee3382c7f029180/flake.nix#L43-L51

srid:
Might as well want to expose `modifier`. For https://github.com/srid/tailwind-haskell/blob/0fb8a18b0e770bafc17521836658f31c56e6dfdb/flake.nix#L18-L27

srid:
@locallycompact's https://horizon-haskell.net/ 's overlay could also be exposed in this manner?

roberth:
> https://github.com/srid/emanote/blob/6f83b857741fb6607716cf97cee3382c7f029180/flake.nix#L43-L51

I think these lines could be turned into a small module that can be imported at the project level.

```
      flake.haskellFlakeProjectModules.default =
        ... those lines ...;
      perSystem = { pkgs, system, inputs', self', ... }: {
        haskellProjects.default = {
          imports = [ self.haskellFlakeProjectModules.default ];
```

srid:
Interesting idea. What happens if two modules define the same option, say `buildTools`? Merge, or replace, or ?

roberth:
It's determined by the option `type`.
`attrsOf` and `lazyAttrsOf` will merge; `listOf` appends. Types like `package` and `str` will error out.
`mkForce` and `mkOverride` replace.

srid:
I suppose for `overrides` we ought to create a custom type with `composeExtensions`-based merge semantics, because the default semantics for `types.functions` just calls the functions and merges the results (attrsets): https://github.com/NixOS/nixpkgs/blob/cd71c681ca4e49de4f65e6a45f686f11f3e1b7d3/lib/types.nix#L620-L621

srid:
It just occurred to me that this approach (`imports`) can also solve the problem of decoupling the hardcoded `buildTools`, so that `imports = [ inputs.haskell-flake.defaultSettings ];` for example could set the `buildTools` defaults (and the user can ignore them if they want to start pristine).

srid:
Example:

- Using: https://github.com/srid/haskell-template/pull/85
- Publishing: https://github.com/srid/nixpkgs-140774-workaround

PR #61, opened 2023-02-05
Add `devShell` submodule; allow disabling dev environment

### LINKED ISSUE #37: Option to disable the development environment

srid:
Resolves #37 (to be squash merged)

- Group `buildTools` (renamed to `tools`), `hlsCheck` and `hlintCheck` under the `devShell` submodule option
- Allow disabling them all using `devShell.enable = false;` (useful if you want haskell-flake to produce _just_ the package outputs).

cc @Kranzes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Revision history for haskell-flake
 
+## `master` branch
+
+- #37: Group `buildTools` (renamed to `tools`), `hlsCheck` and `hlintCheck` under the `devShell` submodule option; and allow disabling them all using `devShell.enable = false;` (useful if you want haskell-flake to produce just the package outputs).
+
 ## 0.1.0
 
 - Initial release
--- a/example/flake.nix
+++ b/example/flake.nix
@@ -12,12 +12,14 @@
         haskellProjects.default = {
           packages = {
             # You can add more than one local package here.
-            example.root = ./.;  # Assumes ./example.cabal
+            example.root = ./.; # Assumes ./example.cabal
           };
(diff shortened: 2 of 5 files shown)


srid:
Aside: regarding `defaultBuildTools`, perhaps that can be decoupled into a separate [haskell-flake module](https://github.com/srid/haskell-flake/issues/12#issuecomment-1334406956), having the user explicitly do:

```nix
haskellProjects.foo.imports = [
  inputs.haskell-flake.defaults
];
```

Less magic this way, as suggested by @maralorn

maralorn:
Nice. I didn’t know that there was a trick to null some of those default tools. Having some tools in there by default is fine. You shouldn’t change that.

I would just suggest that you list which tools those are in the example and in the option description. Also explain how to null them in the option description, probably best via the example value for the option.

Kranzes:
Good stuff Srid, I remember I suggested you to have it this way months ago and you said that it defeats the whole purpose of the project which I disagreed with. It's good that now I won't be forced to having a devshell when sometimes all I want is to package something with a better interface for cabal2nix.

Issue #52, opened 2023-01-05
Add `haskellProjects.foo.finalPackages` to expose the final package set


srid:
We want to expose the following as an option:

https://github.com/srid/haskell-flake/blob/54334cfae9bbb73732bbb1437260017044f68d0b/flake-module.nix#L188

This will be useful here: https://github.com/srid/ema-template/pull/45/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0L63

Where instead of using `lib.getExe pkgs.haskellPackages.ghcid` (incorrect, because ghcid was overriden) I can instead accurately use `lib.getExe config.haskellPackages.main.finalPackages.ghcid`.

roberth:
`perSystem.haskellProjects.<name>.haskellPackages` would have been a nice name, but that's [taken](https://flake.parts/options/haskell-flake.html#opt-perSystem.haskellProjects._name_.haskellPackages).

How about `haskellProject.<name>.finalPackages`?

You'll have to move the overlay logic from the `perSystem` level to the `haskellProjects.<name>` level to make that work, but I think that's a nice place for that logic to be anyway.

srid:
Sure, `finalPackages` is better. 

But, I'm questioning the naming of `haskellPackages` now. Should we rename it to something else, like `basePackages`? The idea being:

```
basePackages --[ source-overrides; overrides ]--> finalPackages
```

roberth:
Sounds good to me!

srid:
How about this API, where we group overlay-creating settings into a submodule called `packageSet`?

```nix
haskellProjects.foo = {
  devShell.enable = false;
  packageSet = {
    # These settings are applied in that order they are shown here.
    base = pkgs.haskell.packages.ghc944;
    local = {
      mypkg.root = ./.;  # Assumes mypkg.cabal
    };
    source-overrides = { .. }; # Overrides by direct path to source
    overrides = self: super:  { .. };  # Raw Haskell overlay
  };
```

Users can then use `config.haskellPackages.foo.packageSet.final` to access all the packages.

Does it make sense, from API documentation/comprehension/simplicity standpoint, to group the overlay related options like this? Or is it complex? I have a tendency introduce unnecessary abstractions, so it would be good to get some feedback here.

maralorn:
Honestly: I think most interfaces are fine if they are well documented.

[...]

PR #63, opened 2023-02-06
Refactor: move impl to inner config

srid:
Ref: https://github.com/srid/haskell-flake/issues/52#issuecomment-1371579338

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -181,95 +181,88 @@ in
                         touch $out
                       '';
                   projectKey = name;
-                  project =
+                  localPackagesOverlay = self: _:
                     let
-                      localPackagesOverlay = self: _:
-                        let
-                          fromSdist = self.buildFromCabalSdist or (builtins.trace "Your version of Nixpkgs does not support hs.buildFromCabalSdist yet." (pkg: pkg));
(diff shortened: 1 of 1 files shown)


flake-module.nix
```diff
@@ -148,8 +148,17 @@ in
...
+                type = types.attrsOf types.raw;
+                description = ''
+                  The flake outputs generated for this project.
+
+                  This is an internal option, not meant to be set by the user.
+                '';
+              };
             };
-          };
+            config = import ./haskell-project.nix (args // { inherit self pkgs; });
```
srid, line 160: @roberth Is there a better way of doing this (instead of the `//`, which seems like kind of a hack)?

srid:
Will add follow-up changes latter.

flake-module.nix
roberth: Yeah, I think we can use module machinery for this, and we probably want that for those little modules with haskell-flake overrides anyway.
 - Above, use `projectSubmodule = submoduleWith { specialArgs = { inherit pkgs self; }; modules = [ ..... ] }`. `specialArgs` should be fine here, as we don't think we want modules to affect those parameters (not 100% sure about `pkgs`, but that can be improved later if necessary)
 - Use `imports = [ ./haskell-project.nix ];` here

PR #65, opened 2023-02-06
Remove `hlintCheck` in favour of treefmt-nix

### LINKED ISSUE #64: Remove `hlintCheck`
treefmt can do this: https://github.com/srid/haskell-template/pull/83/files

And hlint does not need to be run in the devShell, which is all the more reason to decouple it from haskell-flake. haskell-flake should perhaps have a 'recommendations' section in the README, advising people to use treefmt-nix if they want goodies like auto-formatting, hlint, etc.

srid:
Resolves #64


srid:
I think `hlsCheck` is still integral to haskell-flake, because this project is aimed to create a good basic Haskell development environment that works. hlint is quite tangential in that regard; so it can be handled by treefmt.

PR #67, opened 2023-02-07
Make `overrides` an overlay type rather than a `functionTo`

srid:
A step towards #12

- [x] proof of concept https://github.com/srid/haskell-template/pull/85
- [x] new overlay option type
  - [x] make it mergeable
  - [x] finish mkOption implementation (`loc`, etc.)
- [x] test mergeability of other options (`devShell.tools`, `source-overrides`, etc.)
- [ ] how to publish these modules?[^r]
- [x] docs & changelog
- [x] proper tests

[^r]: From @roberth on [Matrix](https://matrix.to/#/!gcrYWdPsIUOFpXFDHB:matrix.org/$ZOYhSXRWLJne7vr9Wz0FkZV27eOzrSa9dRP8LALV2mM?via=matrix.org&via=nixos.dev&via=gaze.systems):
    > I think what's missing from the picture is a flake attribute for publishing those modules, and setting up the module arguments for those modules so that they don't have to rely on variables from the lexical scope. The lexical scope is too specific to the local project, not including things like the consuming project's pkgs and whatnot [..] oh and for publishing a module, you can use `types.deferredModule`, which is takes modules, turns them into a single module, but does cause them to be evaluated; returning a module syntax instead of an evaluated configuration

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -114,7 +114,18 @@ in
                         # returns an attrset.
                         check = lib.isFunction;
                         merge = _loc: defs:
-                          lib.composeManyExtensions (map (x: x.value) defs);
+                          let
+                            logWarning =
+                              if builtins.length defs > 1
+                              then builtins.trace "WARNING[haskel-flake]: Multiple haskell overlays are applied in arbitrary order." null
+                              else null;
+                            overlays =
+                              map (x: x.value)
+                                (builtins.seq
+                                  logWarning
+                                  defs);
+                          in
+                          lib.composeManyExtensions overlays;
                       };
                     in
                     mkOption {
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -101,22 +101,38 @@ in
                   };
                   overrides =
                     let
-                      haskellOverlay = types.mkOptionType {
+                      # WARNING: While the deterministic, it is not determined
+                      # by the user. Thus overlays may be applied in unexpected
+                      # order.
+                      # We need: https://github.com/NixOS/nixpkgs/issues/215486
+                      haskellOverlayType = types.mkOptionType {
                         name = "haskellOverlay";
-                        description = "Haskell overlay function";
+                        description = "An Haskell overlay function";
                         descriptionClass = "noun";
+                        # NOTE: This check is not exhaustive, as there is no way
+                        # to check that the function takes two arguments, and
+                        # returns an attrset.
                         check = lib.isFunction;
-                        merge = loc: defs:
-                          # TODO: What to do with loc?
+                        merge = _loc: defs:
                           lib.composeManyExtensions (map (x: x.value) defs);
                       };
                     in
                     mkOption {
-                      type = haskellOverlay;
+                      type = haskellOverlayType;
                       description = ''
-                        Overrides for the Cabal project
+                        Cabal package overrides for this Haskell project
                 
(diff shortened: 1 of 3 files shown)


flake-module.nix
```diff
@@ -71,7 +71,8 @@ in
               };
             };
           };
-          projectSubmodule = types.submodule (args@{ name, config, lib, ... }: {
+          submoduleWithImports = mod: types.submoduleWith { modules = lib.toList mod; };
+          projectSubmodule = submoduleWithImports (args@{ name, config, lib, ... }: {
```
roberth, line 75: Done in #68

flake-module.nix
```diff
@@ -95,16 +96,28 @@ in
...
-                  Overrides for the Cabal project
+              overrides =
+                let
+                  haskellOverlay = types.mkOptionType {
+                    name = "haskellOverlay";
+                    description = "Haskell overlay function";
+                    descriptionClass = "noun";
+                    check = lib.isFunction;
+                    merge = loc: defs:
+                      # TODO: What to do with loc?
```
roberth, line 107: `loc` is for reporting the definition locations when doing type checking.
I don't think there's a function we could reuse for this, but Nixpkgs `lib/types.nix` has "example" code to work from.
The reason we can't reuse anything is because the interaction between type checking and laziness is quite subtle. The module system hasn't elaborated all the abstractions that might make this stuff obvious, and I don't know if that can be done at all and without making it slow and confusing.

In this case, I don't think we should use `composeManyExtensions`, because we can't really check the function bodies before know what the arguments are.
Another possible issue is that the ordering of the overlays matters, but is quite arbitrary in the module system. Ideally, modules are structured in such a way that the order in which the modules are processed does not matter, but for list and some overlays it does. We might want to detect that; see https://github.com/NixOS/nixpkgs/issues/215486

So this type is a bit of a rabbit hole. We'll want to go for incremental improvements. No checking at all is a valid first solution.

srid: > In this case, I don't think we should use `composeManyExtensions`, because we can't really check the function bodies before know what the arguments are.

But this is not important, right? The error message may not be helpful, but an error is thrown nonetheless, forcing the user to specify the overlay correctly.

> Another possible issue is that the ordering of the overlays matters

Is this blocker for implementing composable overlays that can be `import`'ed, then? Or, is the order not deterministic for simpler cases. Like take this one,

```nix
{ 
  haskellProjects.foo = {
    imports = [ inputs.myCompanyCommon.haskellFlakeProjectDeps; ];
    overrides = self: super:  { /* Local project overrides go here */ };
  };
};
```

Here, I would imagine the global dependency overries from `myCompanyCommon` repo to apply _first_ before the project `overrides` are applied, right? 

If the order is deterministic, I was thinking we could note the limitation of arbitrary order in the option docs.

roberth: The ordering of the definitions passed to the merge function is deterministic in the sense that it only depends on the modules that are loaded and the implementation of the modules system, but I can't tell you what the order would be. It depend on details about how the import closure is computed and when and how often the module list is reversed. Also I don't want to know because it's a better idea for scalability to avoid depending on this implementation detail. If you move an override to another module, and the code breaks, wouldn't that be really really bad? It'd be unreliable and it stops you from refactoring the code, so that will become a mess.

I think we can just document it for now.

srid: By the way, even without this PR this problem remains - attrset merging (which functionTo relies on) too uses arbitrary order, but there it is not even using the overlay composition.

PR #68, opened 2023-02-08
Expose `finalPackages`

roberth:
#52

Access to `finalPackages` was useful for some troubleshooting.
This also adds the docs to CI and adds a script to preview the docs locally.

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -71,7 +71,7 @@ in
               };
             };
           };
-          projectSubmodule = types.submoduleWith { specialArgs = { inherit pkgs self; }; modules = [ ./haskell-project.nix (args@{ name, config, lib, ... }:
+          projectSubmodule = types.submoduleWith { specialArgs = { inherit pkgs self; }; modules = [ ./haskell-project.nix
           {
             options = {
               haskellPackages = mkOption {
@@ -151,7 +151,7 @@ in
                 internal = true;
               };
             };
-          }) ]; };
+          } ]; };
         in
         {
           options.haskellProjects = mkOption {
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -71,7 +71,27 @@ in
               };
             };
           };
-          projectSubmodule = types.submodule (args@{ name, config, lib, ... }: {
+          projectSubmodule = types.submodule (args@{ name, config, lib, ... }:
+          let
+            localPackagesOverlay = self: _:
+              let
+                fromSdist = self.buildFromCabalSdist or (builtins.trace "Your version of Nixpkgs does not support hs.buildFromCabalSdist yet." (pkg: pkg));
+                filterSrc = name: src: lib.cleanSourceWith { inherit src name; filter = path: type: true; };
+              in
+              lib.mapAttrs
+                (name: value:
+                  let
+                    # callCabal2nix does not need a filtered source. It will
+                    # only pick out the cabal and/or hpack file.
+                    pkgProto = self.callCabal2nix name value.root { };
+                    pkgFiltered = pkgs.haskell.lib.overrideSrc pkgProto {
+                      src = filterSrc name value.root;
+                    };
+                  in
+                  fromSdist pkgFiltered)
+                config.packages;
+          in
+          {
             options = {
               haskellPackages = mkOption {
                 type = types.attrsOf raw;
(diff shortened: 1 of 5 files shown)


srid:
@roberth Why does `finalOverlay` use `default` (in `options`) to define the value, unlike [`outputs`](https://github.com/srid/haskell-flake/pull/63/files#diff-0549e7a291dba4b0c8148bb5c499a14bb7c1cd0aa61e74e31623467004ad2b6fR149) which uses `config` to do it? What is the difference between the two, in general?

flake-module.nix
```diff
@@ -133,6 +153,34 @@ in
...
+                description = ''
+                  The final package set, based `haskellPackages` plus
+                  the additions and overrides specified in the other options.
+                '';
+              };
+              finalOverlay = mkOption {
+                type = types.raw;
+                readOnly = true;
+                internal = true;
+                default = lib.composeManyExtensions [
```
srid, line 173: @roberth Why do `finalPackages` and `finalOverlay` use `default` (in `options`) to define their values, unlike [`outputs`](https://github.com/srid/haskell-flake/pull/63/files#diff-0549e7a291dba4b0c8148bb5c499a14bb7c1cd0aa61e74e31623467004ad2b6fR149) which uses `config` to do it? What is the difference between the two, in general?

roberth: The difference is the `mkOverride` priority that is assigned to the definition. In case of a `readOnly = true` option, it doesn't really matter which one you use, because only one definition is allowed anyway, making override priorities irrelevant.

I think placing definitions in `config` is a good habit though, so I'll change this.

srid: > I think placing definitions in `config` is a good habit though, so I'll change this.

Cool; it would also minimize the diff in this PR, I imagine, since all `config` stuff lives in `haskell-project.nix`.

runtest.sh
```diff
@@ -15,4 +15,10 @@ echo "\n||| Testing nix flake checks"
 nix --option sandbox false \
     build --override-input haskell-flake path:${FLAKE} -L .#check
 
+echo "\n||| Testing docs"
+nix --option sandbox false \
+    build --override-input haskell-flake path:${FLAKE} \
+    -L --show-trace \
+    github:hercules-ci/flake.parts-website#checks.x86_64-linux.linkcheck
```
srid, line 22: I'd remove `-L --show-trace` since they are not used in the commands above (or is there a reason for their inclusion, aside from debugging purposes?). Also, must sandbox be disabled for this build?

roberth: The sandbox was disabled in an earlier command, so I assumed that you run this in a setup where that's necessary. This build works fine in the sandbox, but I've only ever built it in a nice NixOS-based sandbox.

The verbosity is helpful when this runs in CI. You'll only look at it when it fails, and then you want all the info. Module system errors tend to need `--show-trace`, at least until we have
 - https://github.com/NixOS/nix/issues/7553

roberth: A proper CI solves this problem in the UI, but admittedly consuming an unpinned input like the site here isn't possible yet in Hercules.

roberth: Anyway, I've changed it so that logs only appear when there's a problem. Still not great if the builder hangs, but that's unlikely to be a problem here.

flake-module.nix
roberth: To move it back, I'd have to change `haskell-project.nix` to a module instead of a function to a `config` value. Basically this 
 - https://github.com/srid/haskell-flake/pull/63/files#r1100250488

I guess why not

flake-module.nix
```diff
@@ -71,7 +71,7 @@ in
               };
             };
           };
-          projectSubmodule = types.submodule (args@{ name, config, lib, ... }:
+          projectSubmodule = types.submoduleWith { specialArgs = { inherit pkgs self; }; modules = [ ./haskell-project.nix (args@{ name, config, lib, ... }:
```
roberth, line 74: All this happening on a single line is ugly, but I think almost all of the things that would otherwise be reindented below can move to the `haskell-project.nix` module later. Doing that here would cause a large diff (again).

flake-module.nix
```diff
@@ -133,37 +134,49 @@ in
                   This is an internal option, not meant to be set by the user.
                 '';
               };
+
+              # Derived options
+
+              finalPackages = mkOption {
+                type = types.attrsOf raw;
+                readOnly = true;
```
srid, line 142: It looks like `readonly` has no effect, because I can still set `finalPackages.foo = ./.;` (even `finalPackages = ./.;`) in user flake.nix?

Or does it just silently ignore user values?

Issue #49, opened 2022-12-01
cabal file in project root is not detected


roberth:
Encountered this in https://github.com/hercules-ci/arion

`haskellPathsInDir` only scans subdirectories: "${x}/${x}.cabal"

We should also try `${x}.cabal` (regardless of parent directory name, which is impure; store path or lazy tree).

srid:
[Trying to describe this issue in detail]

The `haskellProjects.foo.packages` option currently uses the [`haskellPathsInDir` function](https://github.com/NixOS/nixpkgs/blob/853288c39d7955d38bfe72a67be1db6b74078287/lib/filesystem.nix#L11-L18) from nipxkgs to determine its default value:

https://github.com/srid/haskell-flake/blob/380bffdba536ffd6379b236b2bf227985113cc5c/flake-module.nix#L119-L122

This works for projects with *multiple* local packages (like https://github.com/srid/haskell-multi-nix) where cabal files live at `*/*.cabal`. However, we want the default value to also work for single-package haskell projects (like https://github.com/srid/haskell-template), such that the three lines [here](https://github.com/srid/haskell-template/blob/12ec81a94f34a37e99591f9853e9816d0a404652/flake.nix#L25-L27) can be removed and the flake still works, because the default will recognize it.

roberth:
`haskellPathsInDir` didn't really work for `hercules-ci-agent` either, because that project has an integration test driver in a sub-subdirectory; depth 2, which, like depth 0, isn't depth 1.

https://github.com/hercules-ci/hercules-ci-agent/blob/40d5243b4ea4ecedbf021b50272dc0937a2ad4df/flake.nix#L293-L296

So my workaround was to use `lib.mkOptionDefault`, to extend the default, which is quite a fragile solution that I think most users won't be able to figure out, and shouldn't have to.

The package was listed in a `cabal.project` file, so if we could support that, I think we could have a fairly robust solution without having to scan deeply, which could be slow in monorepos.

The algorithm would then be roughly: check for `cabal.project` or `stack.yaml` or `*.cabal` in the root and use that to set the `packages` option; if not present, require the user to set that option (or some helper option that sets it, for instance an option that specifies the location of a `cabal.project` file).
[...]

PR #69, opened 2023-02-08
Added templates.flake-only

### LINKED ISSUE #66: "nix flake init -t github:srid/haskell-flake" creates flake.nix, example.cabal, src/Main.hs
Isn't it supposed to only create the flake.nix?

Will throwing an error as the value for example.root resolve this?

https://github.com/srid/haskell-flake/blob/6a56fbd3e4bb7e040b5ac98ba9fe7afcb5e285c0/example/flake.nix#L13

sbh69840:
Closes #66 

Already existing Haskell projects can use this template to avoid creating _example.cabal_ and _src/Main.hs_ during initialization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,14 @@ To use `haskell-flake` in your Haskell projects, run:
 nix flake init -t github:srid/haskell-flake
 ```
 
-This will generate a template Haskell project with a `flake.nix`. If you already have a Haskell project, copy over this `flake.nix` and adjust accordingly.
+This will generate a template Haskell project with a `flake.nix`. If you already have a Haskell project, run:
+
+
+``` nix
+nix flake init -t github:srid/haskell-flake#flake-only
+```
+
+and adjust accordingly.
 
 ### Template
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -14,7 +14,14 @@ To use `haskell-flake` in your Haskell projects, run:
 nix flake init -t github:srid/haskell-flake
 ```
 
-This will generate a template Haskell project with a `flake.nix`. If you already have a Haskell project, copy over this `flake.nix` and adjust accordingly.
+This will generate a template Haskell project with a `flake.nix`. If you already have a Haskell project, run:
+
+
+``` nix
+nix flake init -t github:srid/haskell-flake#flake-only
```
srid, line 21: One problem here is that this will generate a `flake.nix` hardcoded to use `example.cabal`. But resolving #49 would obviate this issue as we no longer have to specify the package name in the flake.

roberth: `example` is not a terrible placeholder.

srid: - `#default` -> flake only
- `#example` -> full example project

PR #75, opened 2023-02-09
Have `packages` option auto-detect single-package projects

### LINKED ISSUE #49: cabal file in project root is not detected

aravindgopall:
Resolves #49

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -1,5 +1,5 @@
 # A flake-parts module for Haskell cabal projects.
-{ self, config, lib, flake-parts-lib, pkgs, ... }:
+{ self, config, lib, flake-parts-lib, ... }:
 
 let
   inherit (flake-parts-lib)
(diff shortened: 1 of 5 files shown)


flake-module.nix
```diff
@@ -10,6 +10,8 @@ let
   inherit (types)
     functionTo
     raw;
+  root-files = builtins.attrNames (builtins.readDir ./.);
+  currentCabalAttrs = builtins.listToAttrs (builtins.map (cabalFile : { name = pkgs.lib.strings.removeSuffix ".cabal" cabalFile; value = builtins.toString ./.;}) (builtins.filter (name: pkgs.lib.strings.hasSuffix ".cabal" name) root-files));
```
srid, line 14: Instead of `pkgs.lib`, you can also just use `lib` (since it is in the function argument; see line 1).

Also,

- Instead of `./.`, I believe you want to use `self`. Because, the former refers to haskell-flake itself, where as the latter (`self`)` refers to the Haskell project on which this module is being uised.
- `value` should be `{ root = self; }` I think.

srid:
Test is failing because of nix bug: 
- https://github.com/NixOS/nix/issues/7263
- https://matrix.to/#/!gcrYWdPsIUOFpXFDHB:matrix.org/$7gvvTC5TAngtrZqDu12QHOj2lF1m4b55K7o2P-mZMy4?via=matrix.org&via=nixos.dev&via=gaze.systems

PR #79, opened 2023-02-10
Add `haskellFlakeProjectModules` option

srid:
Implements https://github.com/srid/haskell-flake/issues/12#issuecomment-1334406956


srid:
This works in https://github.com/srid/haskell-template/pull/85

image

test/flake.nix
```diff
@@ -23,26 +23,23 @@
         inputs.check-flake.flakeModule
       ];
       perSystem = { self', pkgs, ... }: {
+        haskellFlakeProjectModules.default = {
```
srid, line 26: @roberth I must say I'm not happy with this flake-parts module now providing *two* (not one) top-level options -- `haskellProjects` and, now, `haskellFlakeProjectModules`.

A single top-level would make sense. Like:

```nix
{ 
  haskellFlake = {
    projectModules = { .. };
    projects.default = { .. };
  };
}
```

It also makes the option name obvious enough to indicate that it comes from `inputs.haskell-flake` rather than any other inputs (the same cannot be strictly said for `haskellProjects` naming). A similar naming is adopted by other modules like https://github.com/Platonic-Systems/mission-control

wdyt?

srid: OTOH, the current way of naming it `outputs.haskellFlakeProjectModules.${system}.default` appears to follow the unspoken convention of using flat attributes à la `nixosModules`, `darwinModules`, etc.

test/flake.nix
```diff
@@ -23,26 +23,23 @@
         inputs.check-flake.flakeModule
       ];
       perSystem = { self', pkgs, ... }: {
```
roberth, line 25: I don't think using the local `pkgs` is the right thing to do. It should come from the environment where it's imported, so all we can really need to do is `flake.haskellFlakeProjectModules.foo = { pkgs, ... }: { overrides = f pkgs; };`.

If a flake does need to reference packages it defines, it can do so with `getSystem` and the like.

test/flake.nix
roberth: I think we need options in two separate locations anyway. See previous comment https://github.com/srid/haskell-flake/pull/79/files#r1103655549

> flat attributes à la `nixosModules`,

Flat attributes is probably the way to go. I've made an exception for `flakeModule`, which I'm not eager to remove until the `imports` UX is improved. Not sure exactly what can be achieved in that area yet. My previous attempt found too much resistance, so I'll have to find something creative... Anyway, until then, I'd say flat is good.

PR #81, opened 2023-02-11
Fix optional check generation

daanrs:


srid:
@daanrs I believe the current behaviour is correct. `devShell.enable = false;` is expected to disable the checks as well (since HLS is a development tool). Did you have different expectations?

`hlsCheck.enable` flag is also already respected; see https://github.com/srid/haskell-flake/blob/4585d422cd4428e2279e344d7dc84436bab93300/haskell-project.nix#L101-L103

daanrs:
Oh that makes sense, I thought this was just a simple typo. I should've probably elaborated more, sorry! In the case that devShell is enabled but devShell.hlsCheck is disabled I get the following error when running `nix flake check`:
```shell
error: attribute 'checks' missing

       at /nix/store/wlnkbbrpj1cg57n7b7pb80sx8h90191n-source/flake-module.nix:263:63:

          262|                   (_: project:
          263|                     lib.optionalAttrs project.devShell.enable project.outputs.checks)
             |                                                               ^
          264|                   config.haskellProjects;
```

srid:
Ah, yes. Just fixed it in 34641d4.

daanrs:
Thank you!

PR #83, opened 2023-02-16
added shellHook option to devShell

dark-ether:
when i first tried flake parts, i tried to use haskell flake but i needed shellHook so i ended up not using it, however the only problem for me is the lack of a shellHook. I recently thought wait couldn't i just add a shellHook option?

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -69,6 +69,13 @@ in
                   A [check](flake-parts.html#opt-perSystem.checks) to make sure that your IDE will work.
                 '';
               };
+              shellHook = mkOption {
+                default = "";
+                type = types.str;
+                description = ''
+                  Bash statements that are executed by nix develop.
+                '';
+              };
             };
           };
           projectSubmodule = types.submoduleWith {
--- a/haskell-project.nix
+++ b/haskell-project.nix
@@ -65,6 +65,7 @@ in
             (name: p."${name}")
             (lib.attrNames config.packages);
         withHoogle = true;
+        shellHook = config.devShell.shellHook;
       };
     in
     {


srid:
What if solve this in a more _general_ fashion? ie., by supporting _any_ `mkShell` attribute argument, not just `shellHook`?

For examlpe, we could add a `mkShellArgs` option and use it like:

```nix
haskellProjects.default = {
  devShell = {
    mkShellArgs = {
      shellHook = "echo Hello;";
      nativeBuildInputs = [ myScript ];
      SOME_ENV_VAR = "some":
    };
   };
}
```

cc @roberth (feels like poor man's substitute for a proper module system for devshells)

Incidentally, this will obviate the `installToDevShell` function of https://github.com/Platonic-Systems/mission-control/blob/feb06872ac4dc977f70f6388c87d36fc3c3c3693/nix/flake-module.nix#L78-L91 replacing it with a `mkShellAttrs` option that we can pass as-is to `haskell-flake`'s devShell.

srid:
What if we solve this in a more _general_ fashion? ie., by supporting _any_ `mkShell` attribute argument, not just `shellHook`?

For examlpe, we could add a `mkShellArgs` option and use it like:

```nix
haskellProjects.default = {
  devShell = {
    mkShellArgs = {
      shellHook = "echo Hello;";
      nativeBuildInputs = [ myScript ];
      SOME_ENV_VAR = "some";
    };
   };
}
```

cc @roberth and @blaggacao (feels like poor man's substitute for a proper module system for devshells)

Incidentally, this will obviate the `installToDevShell` function of https://github.com/Platonic-Systems/mission-control/blob/feb06872ac4dc977f70f6388c87d36fc3c3c3693/nix/flake-module.nix#L78-L91 replacing it with a `mkShellAttrs` option that we can pass as-is to `haskell-flake`'s devShell.

roberth:
I think @srid's suggestion is good, but I'd also like to standardize on something that isn't necessarily the same as `mkShell` arguments, so that alternative implementations of the shell idea can exist, without being a hack on top of a hack.
Perhaps something like:

```
{
  env.SOME_ENV_VAR = "some";
  stdenvAttrs.buildInputs = [ pkgs.zlib ];
  environmentPackages = [ pkgs.example-tool ];
}
```

Doing it this way allows a shell implementation to avoid loading `stdenv/setup.sh` unless necessary, for example, or it could build a symlink farm instead of a long `PATH`.

`mkShellArgs` seems like a good intermediate solution though.

srid:
@roberth A *separate* flake-parts module just for constructing devShells but in a composable manner? If that module exposed a `submodule` that can be injected into `haskell-flake` and the like, it would be more interesting than my solution. I wonder how merging multiple devShell attrs (say, merging that of `haskell-flake` and that of `mission-control`) would look like ... perhaps something like:

```nix
inputs.devshell-parts.url = "github:hercules-ci/devshell-parts";
...
{
  devShellParts.default = {
    imports = 
      [ config.haskellProjects.default.devShellPart
        config.misson-control.devShellPart
      ];
  };
  devShells.default = config.devShellParts.default;
}
```

?

roberth:
I hadn't thought of actually using `deferredModule` for that before; just tame data.



> EDIT: `config.haskellProjects.default.devShellPart` maybe tricky to implement since we use `shellFor`

It's possible to extract the arguments with something like `((shellFor x).overrideAttrs(prevAttrs: { passthru.prevAttrs = prevAttrs; })).prevAttrs`. Not pretty, but may be an acceptable way out.

dark-ether:
i have to say i wholeheartedly agree that a devShell flake module would be good, but is it in scope for haskell-flake? And more importantly shouldn't we have a specific place to gather all the discussion and ideas for starting a project like this?  

I am onboard to this idea and would contribute to such a module, the first thing i would consider is that there are 2 types of shells development ones and others for debugging the nix build. Also we probably should exhaustively enumerate and consider each feature/option, some that come to mind are 
- hooks, for exiting and entering the shell, i previously implemented one by using trap but that is an ugly hack.
- packages in path
- package specific configuration
i have to say that if by tomorrow anyone doesn't bite the bullet i will try to implement one

srid:
@dark-ether Agreed, 'composable devshells' is an interesting problem to be solved on its own (and thereon *used* by the likes of haskell-flake).

We talked about this briefly in the following channels

- https://app.element.io/#/room/#boom-nix:matrix.org
- https://app.element.io/#/room/#hercules-ci:matrix.org

PR #87, opened 2023-02-21
User-friendly overrides API

aravindgopall:
example usage:

```nix
          easy-overrides = self: super: {
            aeson = {
              overrides = old: {
                broken = false;
                libraryHaskellDepends = old.libraryHaskellDepends ++ [];
              };
              input = super.aeson;
            };
            relude.overrides.doCheck = false;
          };
```

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -119,7 +119,9 @@ in
                       description = " An attribute set with haskell packages and the options.";
                       type = with types; functionTo (functionTo (attrsOf (submodule {
                         options.overrides = mkOption {
-                          type = oneOf [ (attrsOf raw) (functionTo (attrsOf raw)) ];
+                          type = deferredModuleWith {
+                            staticModules = [ ./package-settings.nix ];
+                          };
                           description = "cabal overrides of the package.";
                           default = { };
                         };
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -115,13 +115,17 @@ in
                     default = { };
                   };
                   easy-overrides =
-                    let settingsFormat = pkgs.formats.json { };
-                    in
                     mkOption {
                       description = " An attribute set with haskell packages and the options.";
-                      type = with types; submodule {
-                        freeformType = settingsFormat.type;
-                      };
+                      type = with types; functionTo (functionTo (attrsOf (submodule {
+                        options.overrides = mkOption {
+                          type = oneOf [ (attrsOf raw) (functionTo (attrsOf raw)) ];
+                        };
+                        options.input = mkOption {
+                          type = nullOr package;
+                          default = null;
+                        };
+                      })));
                       default = { };
                     };
                   overrides =
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -114,6 +114,19 @@ in
                     description = ''Package overrides given new source path'';
                     default = { };
                   };
+                  easy-overrides = mkOption {
+                    description = " An attribute set with haskell packages and the options.";
+                    type = with types; attrsOf (submodule {
+                      options = {
+                        profiling = mkOption {
+                          type = bool;
+                        };
+                        disableTests = mkOption {
+                          type = bool;
+                        };
+                      };
+                    });
+                  };
                   overrides =
                     let
                       # WARNING: While the order is deterministic, it is not
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -120,10 +120,13 @@ in
                       type = with types; functionTo (functionTo (attrsOf (submodule {
                         options.overrides = mkOption {
                           type = oneOf [ (attrsOf raw) (functionTo (attrsOf raw)) ];
+                          description = "cabal overrides of the package.";
+                          default = { };
                         };
                         options.input = mkOption {
                           type = nullOr package;
                           default = null;
+                          description = "derivation of the package.";
                         };
                       })));
                       default = _: _: { };
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -114,29 +114,16 @@ in
                     description = ''Package overrides given new source path'';
                     default = { };
                   };
-                  easy-overrides = mkOption {
-                    description = " An attribute set with haskell packages and the options.";
-                    type = with types; attrsOf (submodule {
-                      options = {
-                        profiling = mkOption {
-                          type = bool;
-                          default = false;
-                        };
-                        disableTests = mkOption {
-                          type = bool;
-                          default = false;
-                        };
-                        doJailbreak = mkOption {
-                          type = bool;
-                          default = false;
-                        };
-                        unmarkBroken = mkOption {
-                          type = bool;
-                          default = false;
-                        };
+                  easy-overrides =
+                    let settingsFormat = pkgs.formats.json { };
+                    in
+                    mkOption {
+                      description = " An attribute set with haskell packages and the options.";
+                      type = with types; submodule {
+                        freeformType = settingsFormat.type;
                       };
-                    });
-                  };
+                      default = { };
+                    };
                   overrides =
                     let
                       # WARNING: While the order is deterministic, it is not
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -120,9 +120,19 @@ in
                       options = {
                         profiling = mkOption {
                           type = bool;
+                          default = false;
                         };
                         disableTests = mkOption {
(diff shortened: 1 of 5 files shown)


haskell-project.nix
```diff
@@ -66,6 +66,16 @@ in
             (lib.attrNames config.packages);
         withHoogle = true;
       };
+      modifyoverrides = s: name: value:
+        pkgs.haskell.lib.compose.overrideCabal
+          (drv: {
+            doCheck = ! value.disableTests;
+            enableLibraryProfiling = value.profiling;
+            enableExecutableProfiling = value.profiling;
```
roberth, line 74: You might structure this in RFC42 style, so that knowledge about `overrideCabal` is transferable between it and haskell-flake.

flake-module.nix
```diff
@@ -99,6 +99,19 @@ in
                     description = ''Package overrides given new source path'';
                     default = { };
                   };
+                  easy-overrides = mkOption {
+                    description = " An attribute set with haskell packages and the options.";
+                    type = with types; attrsOf (submodule {
```
roberth, line 104:
```suggestion
                    type = with types; attrsOf (deferredModulesWith { staticModules = [ { old, ... }: {
```

While this is not a trivial change, it would allow the `old` parameter to be passed in as a module argument. The attributes in `old` could be used as part of the option default definition, bridging the gap between overrides and options.

haskell-project.nix
srid: By 'RFC42' do you mean https://github.com/NixOS/rfcs/pull/42 ?

flake-module.nix
```diff
@@ -99,6 +99,19 @@ in
                     description = ''Package overrides given new source path'';
                     default = { };
                   };
+                  easy-overrides = mkOption {
```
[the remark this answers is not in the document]
srid: Or just `packageSettings`?

roberth:
My understanding was that this PR makes the overrides more composable and documented, while still relying on the existing [perSystem.haskellFlakeProjectModules](https://flake.parts/options/haskell-flake.html#opt-perSystem.haskellFlakeProjectModules) to distribute the settings together with the source overrides.
The possibility of distributing just the overrides is just a side effect of the suggestion I made, but doesn't need to be used.

maralorn:
I like this idea. I am just uncertain about naming and the exact options. For me personally I guess `cabalOverrides` would be a straight forward name, but `packageSettings` also sounds good.

Also wondering how to deal with overrides which need to access the `old` attributes.

I second the idea that it would be great if we could expose all of `overrideCabal.`

I wonder if the source overrides you describe obviate the old source-overrides setting. Maybe we can just replace it?

flake-module.nix
srid: IIUC, this would give us something like:

```nix
{
  easy-overrides = {
    aeson = old: {
      enableLibraryProfiling = ! old.enableLibraryProfiling;
    };
}
```

right? Thus solving @maralorn 's "Also wondering how to deal with overrides which need to access the old attributes." question below.

srid: As long as the resulting API is not complex ...

srid:
I just noticed something. If we do:

```nix
{
  easy-overrides.aeson = {
    patches = [ ./mypatch.patch ];
  };
}
```

This would _delete_ the existing patches, which is generally not what the user wants (they want to [append](https://github.com/NixOS/nixpkgs/blob/ffdae3d38630564a8578143a2e66353f3b44f291/pkgs/development/haskell-modules/lib/compose.nix#LL203C3-L203C85) it usually).

---

How can we resolve this confusion? Perhaps we can shift all cabal overrides to a submodule:

```nix
{
  easy-overrides.aeson = {
    cabalOverrides = {
      doCheck = false;
      # Not used typically
      # patches = [ ./mypatch.patch ];
    };
    appendPatches = [ ./mypatch.patch ];
  };
}
```

(But I don't like the extra indirection, TBH)

This has the added advantage of separating out the source override from cabal overrides:

```nix
{
  easy-overrides.aeson = {
    cabalOverrides = {
      doCheck = false;
    };
    source.version = "1.5.1";  # TODO: How do we make `source` an enumOf hackage version, drv, path?
  };
}
```

srid:
@aravindgopall To fix the CI, merge the latest changes from `master`.

flake-module.nix
aravindgopall: @roberth Can you help me to understand to advantage of using `deferredModuleWith`, rather than something like below (which is working).

```nix
                  easy-overrides =
                    let settingsFormat = pkgs.formats.json { };
                    in
                    mkOption {
                      description = " An attribute set with haskell packages and the options.";
                      type = with types; submodule {
                        freeformType = oneOf [(attrsOf raw) (functionTo (attrsOf raw))];
                      };
                      default = { };
                    };
```

flake-module.nix
```diff
@@ -114,6 +114,20 @@ in
                     description = ''Package overrides given new source path'';
                     default = { };
                   };
+                  easy-overrides =
+                    mkOption {
+                      description = " An attribute set with haskell packages and the options.";
+                      type = with types; functionTo (functionTo (attrsOf (submodule {
```
aravindgopall, line 120: @roberth can you help here if there is a better way to implement this type.

flake-module.nix
roberth: `submodule` returns the computed fixpoint `config`, whereas `deferredModule` lets the module be a module that's reusable, as a function.

Also `freeformType` doesn't really support anything other than an attrset at its root. It would never get to see the module function.

flake-module.nix
roberth: See thread https://github.com/srid/haskell-flake/pull/87#discussion_r1113327459

flake-module.nix
roberth: It's hard to communicate the implementation pattern, so I've made an example here 
 - https://github.com/srid/haskell-flake/pull/96
 
I think the main benefit is that you get the module syntax and module composition for free. It will be easy to define merge logic, which would be annoying to write with a custom type.

I think adding a `freeformType` to the package settings module would be a good idea. My example didn't include that, but it would allow undeclared parameters to be set, so that users won't be held back when they need to set something that's not explicitly added in haskell-flake yet.

aravindgopall: Tried with freeform already and it was working using `evalModules`, but just curious to understand the advantage of one over the other. Thanks for the explanation and the PR.

aravindgopall: > Also freeformType doesn't really support anything other than an attrset at its root. It would never get to see the module function.
But in the above shared example using `oneOf`, using functionTo which is not really a attrset at the root level but this is working with freeformType.

srid: > [..] it would allow undeclared parameters to be set, so that users won't be held back when they need to set something that's not explicitly added in haskell-flake yet.

I actually think that using explicit options (as in #96), rather than a freeform attrset, is beneficial in that regard.

srid: ... unless we get as useful/simpler error messages when using freeformType (I'm not sure if we can).

srid:
If we were to extend the prototype #96 to allow passing in the `self` and `super` (in addition to, thus, being able to set their source/derivation, such as to allow us to be able to get rid of `source-overrides`), viz.:

```nix
{
  packageSettings = { self, super, ... }: {
    aeson.flags = { old, ... }: { enableLibraryProfiling = ! old.enableLibraryProfiling; };
    aeson.input = self.callHackage "aeson" "1.5.0.2" {};
  };
}
```

... then there would be *two* levels of `deferredModuleWith`, would it not?

It would also allow us to `import` package settings from other places, viz.:

```nix
{
  packageSettings.imports = [ inputs.my-company-deps.haskellFlakePackageSettings.default ];
}
```

srid:
Let's test this PR out in https://github.com/nammayatri/nammayatri/pull/14/files#diff-2a1dbd143fde4decc1da839b68cb6d875287d0bacb43e6d848ea14e6f6a6e6d2

srid:
A big problem with using `functionTo (functionTo` as a type for easy-overrides is that it cannot be merged. We need to fix this. I also realized that you can't create a `attrsOf (deferredModuleWith { ...`.

I have to rethink this.

aravindgopall:
> 

We have to create a new type like we did for overlay and it can take attrset (key - pacageName and value is overrides) as input.

PR #92, opened 2023-02-23
Add `devShell.mkShellArgs`

srid:
Resolves #83

Details: https://github.com/srid/haskell-flake/pull/83#issuecomment-1433427743

(We need `shellHook` for generating .cabal from package.yaml, so HLS will work.)


srid:
This is just an intermediate solution until we have composable devshells.

PR #96, opened 2023-02-26
Proof of concept packageSettings with deferredModule

roberth:


test/flake.nix
```diff
@@ -32,6 +32,9 @@
             # Custom library overrides (here, "foo" comes from a flake input)
             foo = self.callCabal2nix "foo" (inputs.haskell-multi-nix + /foo) { };
           };
+          packageSettings = {
+            haskell-flake-test.broken = true;
```
roberth, line 36: Alternatively, this may reference other values in `old`:
```suggestion
            haskell-flake-test = { old, ... }: { broken = ! old.broken; };
```

It allows somewhat advanced logic to be expressed modularly, without requiring users to define a module _function_ most of the time, as you can see in the "before" of the suggestion.

srid:
Closing per https://github.com/srid/haskell-flake/pull/87#issuecomment-1450286992

Issue #82, opened 2023-02-14
Facilitate predictable overlay merging


srid:
Background: https://github.com/srid/haskell-flake/pull/67#discussion_r1101835307

In the below imports, the overlays are applied in *reverse* order:

```nix
            imports = [
              {
                overrides = self: super: {
                  aeson = super.aeson-backup;
                };
              }
              {
                overrides = self: super: {
                  aeson-backup = super.aeson;
                  aeson = null;
                };
              }
            ]
```

The last overlay is applied first, and then the first one is applied next. This is how it is as of https://github.com/NixOS/nixpkgs/commit/bbf77421ac51a7c93f5f0f760da99e4dbce614fa - but there is no guarantee that this order will not change.

Currently when multiple overlays are in use, `haskell-flake` will print this warning:

```
trace: WARNING[haskell-flake]: Multiple haskell overlays are applied in arbitrary order.
```

But still, this can be surprising and annoying to deal with.

We want something like https://github.com/NixOS/nixpkgs/issues/215486 - or at least something to allow the current project overrides take priority (even if the overlays of dependencies themselves, when `import`ed, are applied in arbitrary order).

srid:
For this, however, the project overlay (the one below imports) are applied last. As if the module system applies the `imports` first, before merging the options in the module doing the importing.

```nix
            imports = [
              {
                overrides = self: super: {
                  aeson-backup = super.aeson;
                  aeson = null;
                };
              }
            ];
            overrides = self: super: {
              aeson = super.aeson-backup;
            };
```

PR #98, opened 2023-02-28
DRAFT Add `packageSettings`

srid:
This is a follow-up to https://github.com/srid/haskell-flake/pull/87#issuecomment-1448903620 (and closes #87 and closes #96).

The module API is slightly diferrent here. While it still uses `deferredModule` (like #96), the `config` implementation of this PR expands directly to `overrides`, thus we can reuse the merge semantics of `overrides` without worrying about writing one for `packageSettings`. We also add `input.hackageVer` as an easier way of doing `self.callHackage ...`.

- [ ] Assess how overlay composition order is affected as a result of this
- [ ] API is slightly clunky owing to what appears to be limitations of the module system (see diff comments)
- [ ] Deprecate/remove `source-overrides` (since the `input` option can do that and more)

### DIFF
--- a/nix/package-settings.nix
+++ b/nix/package-settings.nix
@@ -47,7 +47,7 @@
                       default = null;
                       description = "Whether to run tests";
                     };
-                    doJailbreak = lib.mkOption {
+                    jailbreak = lib.mkOption {
                       type = lib.types.nullOr lib.types.bool;
                       default = null;
                       description = "Whether to disable version bounds";
--- a/nix/package-settings.nix
+++ b/nix/package-settings.nix
@@ -67,6 +67,11 @@
                       default = null;
                       description = "Patches to apply";
                     };
+                    enableSeparateBinOutput = lib.mkOption {
+                      type = lib.types.nullOr lib.types.bool;
+                      default = null;
+                      description = "Whether to enable separate bin output";
+                    };
                   };
                 })
               ];
--- a/nix/package-settings.nix
+++ b/nix/package-settings.nix
@@ -57,6 +57,11 @@
                       default = null;
                       description = "Whether to generate documentation";
                     };
+                    patches = lib.mkOption {
+                      type = lib.types.nullOr (lib.types.listOf lib.types.path);
+                      default = null;
+                      description = "Patches to apply";
+                    };
                   };
                 })
               ];
--- a/nix/package-settings.nix
+++ b/nix/package-settings.nix
@@ -52,6 +52,11 @@
                       default = null;
                       description = "Whether to disable version bounds";
                     };
+                    broken = lib.mkOption {
+                      type = lib.types.nullOr lib.types.bool;
+                      default = null;
+                      description = "Whether the package is broken";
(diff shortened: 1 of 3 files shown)


test/flake.nix
```diff
@@ -46,6 +46,15 @@
             # docs.
             foo = self.callCabal2nix "foo" (inputs.haskell-multi-nix + /foo) { };
           };
+          # WIP test
+          packageSettings = {
+            bar.input = { self, super, ... }: {
```
srid, line 51: I'm forced to add `{ self, super, ...}` here, in `bar.input`, rather than at the toplevel where it is more suitable to have. ie., `packageSettings = { self, super, ... }:` - however I have no idea how you would implement a deferredModule whose staticModule has a **dynamic** list of top-level options (ie., attr keys like foo, aeson, etc.). Is this even possible?

nix/package-settings.nix
```diff
@@ -0,0 +1,102 @@
...
+  options = {
+    packageSettings = lib.mkOption {
+      type = lib.types.lazyAttrsOf (lib.types.submodule {
+        options = {
+          input = lib.mkOption {
+            default = { };
+            type = lib.types.deferredModuleWith {
+              staticModules = [
+                ({ lib, self, super, ... }: {
+                  # TODO: What is the best way to do enum type here?
```
srid, line 15: Normally there is `types.oneOf`, but I can't use it here because these are top-level static module options. Is there a better way of doing this?

srid: If we can do the top-level `{self, super, ...}` as indicated in the above comment, I could simply make `input` an `oneOf` in the style of https://github.com/NixOS/nixpkgs/blame/995edc972ad3a1e291ac22d74b9610821357175f/pkgs/development/haskell-modules/lib/compose.nix#L44-L54

srid:
The name `packageSettings` may no longer be appropriate/accurate since this option can also specify custom packages from, say, flake inputs (see https://github.com/nammayatri/nammayatri/pull/37). It is pretty much looking like a friendly wrapper for `overrides`.

srid:
> Assess how overlay composition order is affected as a result of this

Unfortunately, `packageSettings` take priority over `overrides`, irrespective of the order `modules` list is in.

srid:
> > Assess how overlay composition order is affected as a result of this
> 
> Unfortunately, `packageSettings` take priority (cf. #82) over `overrides`, irrespective of the order `modules` list is in.

Treating `packageSettings` as a distinct overlay solves this; because we control the order in `composeManyExtensions`: #99 

Now we have no choice but to look into the semantics of merging of multiple `packageSettings` values.

srid:
> Now we have no choice but to look into the semantics of merging of multiple `packageSettings` values.

Okay, we can't merge these modules (there will be errors like `error: The option `broken' has conflicting definition values`). That sucks.

srid:
Closing per https://github.com/srid/haskell-flake/pull/87#issuecomment-1450286992

Issue #62, opened 2023-02-06
Configurable auto-wiring


srid:
Currently `haskell-flake` implicitly sets `packages`, `devShells` and `checks`. Let's call it "auto wiring". 

Unless upstream [implements it](https://github.com/hercules-ci/flake-parts/issues/114), we can add an `haskellProjects.foo.autoWire` bool option (default is true) that users can disable if they want to "wire" the flake outputs themselves, eg.:

```nix
{
  haskellProjects.foo = {
    ..
    autoWire = false;
  };
  packages.default = config.haskellProjects.foo.output.packages.mypkg;
  devShells.default = config.haskellProjects.foo.output.devShell;
  # Disabling checks temporarily
  checks = config.haskellProjects.foo.output.checks;
}
```

srid:
In the context of #84, here's an alternate API:

```nix
{
  flake.haskellFlake.autoWire = true;
}
```

Note: `flake.haskellFlake.autoWire` operates *only* on the `haskellProject.default` project, and it is enabled by default. Therefore, by default, outputs are auto-wired for the "default" project. But *other* projects are not injected into the flake outputs (in contrast to current behaviour); the user has to manually do that. This can be desirable behaviour as it removes the naming hack (`haskellProjects.foo` -> `packages.foo-whatever`) and forces to user to come up with good name for their package output.

srid:
Or:

```nix
{  # In perSystem
  haskell-flake = {
    projects.default = { .. };
    autoWire = true;
  };
}
```

(Assuming it is possible to wire `flake` outputs from inside of a `perSystem` config .... if it even makes sense 🤔)

PR #102, opened 2023-03-02
Revamp 'outputs'

### LINKED ISSUE #74: Re-evaluate the structure of `outputs`
I don't think we need this option (added in #63). Here's why:

- `finalPackages` already contains `outputs.packages`
- For `outputs.devShells`, we can expose the devshell in `config.haskellPackages.foo.devShell.drv`
- Likewise, the hls check in `outputs.checks` can be exposed in `config.haskellPackages.foo.devShell.hlsCheck.drv`.

I can't think of a reason to retain `outputs` option in its current form.

srid:
Resolves #74

Also a step towards #62

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - New features
   - #63: Add `config.haskellProjects.${name}.outputs` containing all flake outputs for that project.
-    - #??: In addition, `outputs` contains `finalPackages` and `localPackages`.
+    - #102 In addition, `outputs` contains `finalPackages` and `localPackages`.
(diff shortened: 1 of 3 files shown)


srid:
I went with putting `finalOverlay` inside of "outputs" because .. it is technically an *output*. It is still marked as `internal`, though.

srid:
An unintended benefit of doing this: we can now use haskell-flake just to define arbitrary package sets, and use them elsewhere:

https://github.com/nammayatri/common/pull/3

```nix
{
  basePackages = config.haskellProjects.ghc810.outputs.finalPackages;
}
```

Should document this 'recipe' ...

srid:
> 
> Should document this 'recipe' ...

https://haskell.flake.page/package-set

PR #106, opened 2023-03-06
Add a default for `haskellFlakeProjectModules`

### LINKED ISSUE #84: Autowire `haskellFlakeProjectModules`
Let's say that 'qux' depends on 'bar' which depends on 'foo'.

At the moment, bar's `flake.nix` manually wires things up like this:

```nix
      flake.haskellFlakeProjectModules = rec {
        input = { pkgs, ... }: {
          imports = [
            inputs.foo.haskellFlakeProjectModules.output
          ];
        };
        output = { pkgs, lib, ... }: withSystem pkgs.system (ctx@{ config, ... }: {
          # Pass along the dependency overrides to consuming flake.
          imports = [
            input
          ];
          # Pass local packages, as overlay, to the consuming flake.
          source-overrides =
            lib.mapAttrs (name: ks: ks.root)
              config.haskellProjects.default.packages;
        });
      }
```

Notice how 'bar' uses 'foo's module, thus inheriting its overlays and local packages. Thereon, it exports it all, along with its own location packages, for 'qux' to consume from (via importing its `output` module).

The general pattern here is to generate both `input` and `output` modules, thus enabling the user to wire them up to form a *graph* of Haskell packages. **The overall goal here is simplify the use of haskell-flake when managing multiple Haskell packages across multiple repos (as opposed to a mono repo) that depend on each other**.

---

The purpose of this issue is to figure out the best way to minimize this boilerplate while enabling this pattern, and implement it. Here's one approach:

```nix
  perSystem.haskellProjects.default.autoWire.outputs = true;  # Creates 'packages', 'devShells', etc.
  perSystem.haskellProjects.default.autoWire.modules = true;  # Creates 'flake.haskellFlakeProjectModules.{input,output}'
```

The first option would be for #62. I'm not sure if the second option should be in `perSystem` or `flake`. If it has to be in `flake`, we would possibly have something like `flake.haskellFlake.autoWire.modules = true;` - but it feels weird to spread this outside of project config. OTOH, the latter is more agreeable when we consider the fact that autowiring of modules is done only for the 'default' project.

srid:
Resolves #84 

The idea here is to expose the local packages and overrides for reuse in downstream flake. Could well build a network of haskell flakes this way. 

- [x] Prototype
- [x] Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
     - #52: Expose the final package set as `finalPackages`. Rename `haskellPackages`, accordingly, to `basePackages`. Overlays are applied on top of `basePackage` -- using `source-overrides`, `overrides`, `packages` in that order -- to produce `finalPackages`.
     - #68: You can now use `imports` inside of `haskellProjects.<name>` to modularize your Haskell project configuration.
       - #79: `flake.haskellFlakeProjectModules.<name>` option can be used to set and expose your Haskell project modules to other flakes.
-        - #106: Auto-generate the default modules in 'outputs' for reuse in downstream flakes.
+        - #106: Auto-generate the default modules in 'haskellFlakeProjectModules' for reuse in downstream flakes.
       - #67: `overrides` will be combined using `composeManyExtensions`, however their order is arbitrary. This is an experimental feature, and a warning will be logged.
 
 ## 0.1.0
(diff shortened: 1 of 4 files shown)


srid:
Without this PR we must manually wire things up ourselves:  https://github.com/nammayatri/shared-kernel/pull/29/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R42-R66

image

srid:
Since this is a `default`, users can always disable it by setting `flake.haskellProjectFlakeModules = {}` ... however, I'm not sure how they can *override* (say, by adding modules without removing the default).

PR #110, opened 2023-03-09
Autodetect "packages" based on `cabal.project` (and `package.yaml`)

### LINKED ISSUE #76: Default for `packages` option should use `cabal.project` if it exists
... instead of the current `*/*.cabal` (or `*.cabal`) globbing.

> The algorithm would then be roughly: check for `cabal.project` or `stack.yaml` or `*.cabal` in the root and use that to set the `packages` option; if not present, require the user to set that option (or some helper option that sets it, for instance an option that specifies the location of a `cabal.project` file).

An error message should also be provided when the user hasn't set this option and their project has multiple-packages (no top-level .cabal) but without a `cabal.project.`

_Originally posted by @roberth in https://github.com/srid/haskell-flake/issues/49#issuecomment-1423912295_

srid:
Resolves #76 & adds a `projectRoot` option.

This PR doesn't implement support for globs; users are expected to write down each package by hand in `cabal.project`. Latter we can improve this.

- [x] Working Prototype
- [x] Edge cases
  - Sometimes the `cabal.project` can live in a sub directory (eg: `./Backend`). Do we support this case, and if so how? (EDIT: Can just add `projectRoot` option)
- [x] Add tests for the parser; cf. https://www.tweag.io/blog/2022-09-01-unit-test-your-nix-code/
- [x] Refactor, ~~and maybe consider putting this code in external flake?~~
- [x] Friendly error messages
  - [x] On parse failure in particular instruct the user what to do rather than just `throw`ing parse errors
  - [x] If the package path has globs (`*`), print the limitation asking the user to set "packages" manually.
- [x] Documentation

~~With this PR, haskell-flake gets its first flake dependency https://github.com/kanwren/nix-parsec which is *slightly* concerning, because the goal is to keep haskell-flake a thin abstraction on top of nixpkgs without growing to be a behemoth (eg: haskell.nix).~~ (EDIT: Decided to just use `fetchGit` for this dependency, which is only evaluated if the user evaluates the default)


srid:
Implementation works (no globs supported); what's left - refactoring, better error messages, and documentation update.

peterbecich:
This is great, makes it easier to use `haskell-flake`, thanks!

PR #111, opened 2023-03-09
Add `devShell.extraLibraries` option

roberth:
> Extra Haskell libraries available in the shell's environment.

I'm already using a worse implementation of this in hercules-ci-agent to make as `runghc` shebang script work: https://github.com/hercules-ci/hercules-ci-agent/commit/6da8c81312b36e81e0f674780585571998c47d51 (quickly hacked together at release time...)

This is an improved version of that with
 - a test
 - no clobbering of the `extraDependencies` parameter

I've also validated it locally with hercules-ci-agent.

### DIFF
--- a/nix/flake-module.nix
+++ b/nix/flake-module.nix
@@ -63,6 +63,20 @@ in
                   Build tools useful for Haskell development are included by default.
                 '';
               };
+              extraLibraries = mkOption {
+                type = functionTo (types.attrsOf (types.nullOr types.package));
+                description = ''
+                  Extra Haskell libraries available in the shell's environment.
+                  These can be used in the shell's `runghc` and `ghci` for instance.
+
+                  The argument is the Haskell package set.
+
+                  The return type is an attribute set for overridability and syntax, as only the values are used.
+                '';
+                default = hp: { };
+                defaultText = lib.literalExpression "hp: { }";
+                example = lib.literalExpression "hp: { inherit (hp) releaser; }";
+              };
               hlsCheck = mkOption {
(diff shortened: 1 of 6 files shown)


nix/flake-module.nix
```diff
@@ -63,6 +63,20 @@ in
                   Build tools useful for Haskell development are included by default.
                 '';
               };
+              extraLibraries = mkOption {
+                type = functionTo (types.attrsOf (types.nullOr types.package));
```
srid, line 67: When would someone want to use `null` as a value?

roberth: `null` just happens to be usable in places where a package is expected. This is somewhat useful; for example `base` is defined as `null` because it's implicitly included everywhere already.
Using `null` instead of, say, `lib.optionalAttrs` to remove the entire attribute makes some code simpler and lazier and therefore more recursion friendly: if `a.c` is nullable, `a.b` can be evaluated by evaluating subexpression `a`'s attrset, without `c`'s condition, whereas if `c`'s presence is optional, the condition has to be computed as part of subexpression `a` in `a.b`.

srid: I was wondering why would someone want to use `null` as a value here in `extraLibraries` in particular (as an actual use case) - but it sounds like you are saying this type is generally useful in any context; okay.

roberth: Yeah, I don't think anyone will set `base` or `text` or `Cabal` here, but if they do, it should be accepted, as Nixpkgs Haskell does.

Issue #88, opened 2023-02-22
Remove `pkgFiltered`


srid:
contd. from https://github.com/srid/haskell-flake/issues/85#issuecomment-1437429317

@roberth I still don't understand why we need this transformation: https://github.com/srid/haskell-flake/blob/34641d4508c2ad00d1a5ef5fb592f49bfa9e2770/haskell-project.nix#L46-L48

If the code is not needed for any project or has no effect, it should be removed to minimize maintenance.

I'm trying to understand [your comment here](https://github.com/srid/haskell-flake/issues/85#issuecomment-1437483533)

> I think it's supposed to cut the `src` directory from any parent directory contents [..] Maybe there's something different about the situation where I used it

Is there a concrete project example that demonstrates the utility of this code? Without which the project will rebuild unnecessarily? I cannot see it in https://github.com/srid/haskell-multi-nix - so I wonder what I'm missing.

srid:
From [a comment](https://github.com/srid/haskell-flake/issues/85#issuecomment-1438987426) by @ParetoOptimalDev

> I can't tell you how to reproduce it, but I feel like I've ran into this rebuild issue as well. That's not very actionable though except perhaps to encourage a little more thought or nudge me to go try and reproduce it again if I have time.

Yea, it would be good if we can create a reliable repro. I feel uncomfortable having code that I have no understanding of. :-)

roberth:
I've played around with this a bit and couldn't find a reproducer despite a couple of variations of passing the source:
 - path string
 - path value
 - source attrset from `cleanSourceWith`

None of these variations reproduced the problem.

You have my permission to remove the extra filter. If rebuilds reappear, we'll have a reproducer.

PR #114, opened 2023-03-11
Source filtering

roberth:
Prevent rebuilds caused by changes in the parent directory, and/or sibling directories.

Reproduced in practice in `hercules-ci-agent` and easy to reproduce with a new test case.


srid:
@roberth I'm testing this on https://github.com/srid/haskell-multi-nix

Changing `README.md` and then running `nix run` rebuilds "foo" and "bar". Shouldn't this PR have fixed that?

roberth:
Looks ok when I run this:

```
[haskell-multi-nix]$ nix eval .#packages.x86_64-linux.bar.drvPath --override-input haskell-flake github:srid/haskell-flake/source-filter
warning: Git tree '/home/user/h/haskell-multi-nix' is dirty
warning: not writing modified lock file of flake 'git+file:///home/user/h/haskell-multi-nix':
• Updated input 'haskell-flake':
    'github:srid/haskell-flake/86695baa937a2aae6968c739a5493dee8ecde05c' (2023-02-19)
  → 'github:srid/haskell-flake/b05dc0395f69c9d2a560c791ab6c9f4676abff6e' (2023-03-11)
"/nix/store/0zc31pwd71pk7h46p9v2hj00hyl5vwdc-bar-0.1.0.0.drv"

[haskell-multi-nix]$ echo foo >>README.md 

[haskell-multi-nix]$ nix eval .#packages.x86_64-linux.bar.drvPath --override-input haskell-flake github:srid/haskell-flake/source-filter
warning: Git tree '/home/user/h/haskell-multi-nix' is dirty
warning: not writing modified lock file of flake 'git+file:///home/user/h/haskell-multi-nix':
• Updated input 'haskell-flake':
    'github:srid/haskell-flake/86695baa937a2aae6968c739a5493dee8ecde05c' (2023-02-19)
  → 'github:srid/haskell-flake/b05dc0395f69c9d2a560c791ab6c9f4676abff6e' (2023-03-11)
"/nix/store/0zc31pwd71pk7h46p9v2hj00hyl5vwdc-bar-0.1.0.0.drv"
```

Same derivation path, so that's a reused build.

roberth:
My tree was dirty before the experiment, but cleaning and redoing gives the same outcome.

srid:
I tried it on Linux, and the result is the same:

```
srid on pce haskell-multi-nix on  master via λ
❯ nix eval .#packages.x86_64-linux.bar.drvPath --override-input haskell-flake github:srid/haskell-flake/source-filter
warning: not writing modified lock file of flake 'git+file:///home/srid/code/haskell-multi-nix':
• Updated input 'haskell-flake':
    'github:srid/haskell-flake/56d8d9787ad8914bf54c2580a4dbb80adbe38f35' (2023-03-10)
  → 'github:srid/haskell-flake/b05dc0395f69c9d2a560c791ab6c9f4676abff6e' (2023-03-11)
"/nix/store/z4syg8rqq48ybiz10s3jn6ka60j88lh5-bar-0.1.0.0.drv"

srid on pce haskell-multi-nix on  master via λ took 11s
❯ echo foo >>README.md

srid on pce haskell-multi-nix on  master [!] via λ
❯ nix eval .#packages.x86_64-linux.bar.drvPath --override-input haskell-flake github:srid/haskell-flake/source-filter
warning: Git tree '/home/srid/code/haskell-multi-nix' is dirty
warning: not writing modified lock file of flake 'git+file:///home/srid/code/haskell-multi-nix':
• Updated input 'haskell-flake':
    'github:srid/haskell-flake/56d8d9787ad8914bf54c2580a4dbb80adbe38f35' (2023-03-10)
  → 'github:srid/haskell-flake/b05dc0395f69c9d2a560c791ab6c9f4676abff6e' (2023-03-11)
"/nix/store/ws0yd930drz9sm6if34v6lk15k36zd2w-bar-0.1.0.0.drv"

srid on pce haskell-multi-nix on  master [!] via λ took 3s
❯
```

Nix info,

```
❯ nix-info
system: "x86_64-linux", multi-user?: yes, version: nix-env (Nix) 2.13.2, channels(root): "nixos-23.05pre449318.2caf4ef5005", nixpkgs: /nix/var/nix/profiles/per-user/root/channels/nixos
```

What could be different about your environment?

srid:
Strangely, it works for `./test/with-subdir` in this PR. Just not haskell-multi-nix ....

srid:
Oh, the cause was the use of *older* nixpkgs in haskell-multi-nix. Advancing it this way resolves it:

• Updated input 'nixpkgs':
    'github:nixos/nixpkgs/bbf77421ac51a7c93f5f0f760da99e4dbce614fa' (2022-11-09)
  → 'github:nixos/nixpkgs/bb31220cca6d044baa6dc2715b07497a2a7c4bc7' (2022-12-19)

Out of curiosity, what changed in nixpkgs recently (since 11-09) regarding this behaviour?

nix/haskell-project.nix
```diff
@@ -55,14 +58,31 @@ in
...
-            # the original source level.
-            let root = realiseHpack name value.root;
-            in fromSdist (self.callCabal2nix name root { })
+          (name: pkgCfg:
+            lib.pipe pkgCfg
+            [
+              (pkgCfg: self.callCabal2nix name pkgCfg.root { })
+
+              # Avoid rebuilding because of changes in parent directories
+              (setSrc name pkgCfg.root)
```
srid, line 80: What's the rationale behind the use of `lib.cleanSourceWith` somehow not triggering rebuilds that are otherwise triggered?

srid (commented):
@roberth I removed the `srcSrc` call, but the tests still pass (on macOS at least). So the `setSrc`/`cleanSourceWith` part appears to be redundant (reducing this PR to mainly adding a test case to ensure that rebuilds don't happen).

```
❯ g diff
diff --git a/nix/haskell-project.nix b/nix/haskell-project.nix
index 0afc23d..033d5e5 100644
--- a/nix/haskell-project.nix
+++ b/nix/haskell-project.nix
@@ -76,9 +76,6 @@ in
             [
               (pkgCfg: self.callCabal2nix name pkgCfg.root { })
 
-              # Avoid rebuilding because of changes in parent directories
-              (setSrc name pkgCfg.root)
-
               # Make sure all files we use are included in the sdist, as a check
               # for release-worthiness.
               fromSdist

❯ ./runtest.sh 
[..]

||| All tests passed!
```

---

cf. https://github.com/srid/haskell-flake/issues/88

nix/haskell-project.nix
srid: You may want to add a test to reproduce the rebuild such that only the use of `setSrc` can fix it.

srid: > I removed the `srcSrc` call, but the tests still pass (on macOS at least)

Bizarrely, this is not the case with haskell-multi-nix. Your PR *does* prevent rebuild. Using `master` branch of haskell-flake causes rebuilds when README.md changes. What's different about `test/with-subdir` though?

test/with-subdir/flake.nix
```diff
@@ -0,0 +1,35 @@
...
+  # Since there is no flake.lock file (to avoid incongruent haskell-flake
+  # pinning), we must specify revisions for *all* inputs to ensure
+  # reproducibility.
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/bb31220cca6d044baa6dc2715b07497a2a7c4bc7";
+    flake-parts.url = "github:hercules-ci/flake-parts/7c7a8bce3dffe71203dcd4276504d1cb49dfe05f";
+    check-flake.url = "github:srid/check-flake/48a17393ed4fcd523399d6602c283775b5127295";
+
+    haskell-multi-nix.url = "github:srid/haskell-multi-nix/7aed736571714ec12105ec110358998d70d59e34";
+    haskell-multi-nix.flake = false;
```
srid, line 11: This is unused.

test/with-subdir/flake.nix
```diff
@@ -0,0 +1,35 @@
...
+    # specified using --override-input to point to ../. For example, 
+    #   `nix build --override-input haskell-flake ..`
+    haskell-flake = { };
+  };
+  outputs = inputs@{ self, nixpkgs, flake-parts, ... }:
+    flake-parts.lib.mkFlake {
+      inputs = inputs // {
+        # work around https://github.com/NixOS/nix/pull/7796 for nix <2.14
+        self = self // { outPath = ./.; };
+      };
```
srid, line 23: We won't need this workaround if you use `${NIX}` (instead of `nix`) in `runtest.sh`

nix/haskell-project.nix
srid: Looking inside the two `.drv`, I see that the source hash changes, yet the tarball contents themselves are identical:

```
❯ nix run nixpkgs#pkgdiff /nix/store/9qy8hbsbawcw1crmr9nbljr3yyn5skzw-source.tar.gz /nix/store/f2daz0iv6dn8j6iw95s16rw5zn27ca8w-source.tar.gz
reading packages ...
WARNING: different package names ("9qy8hbsbawcw1crmr9nbljr3yyn5skzw" and "f2daz0iv6dn8j6iw95s16rw5zn27ca8w")
comparing packages ...
creating report ...
result: UNCHANGED
report: pkgdiff_reports/9qy8hbsbawcw1crmr9nbljr3yyn5skzw/source_to_source/changes_report.html
```

roberth: > What's the rationale behind the use of `lib.cleanSourceWith` somehow not triggering rebuilds that are otherwise triggered?

`pkgCfg.root` is a value like `"${root}/mypackage"`, so something like `"/nix/store/...-source/mypackage"`, where `/nix/store/...-source` contains everything at the root of the flake.

`cleanSourceWith` does not copy things from the parent directory, so it returns a store path that only contains the relevant directory.

> Looking inside the two `.drv`, I see that the source hash changes, yet the tarball contents themselves are identical:

That means that `cabal sdist` is doing its job. You'll need the `src.src` - the source of the sdist build - to be identical too. Otherwise, the sdist builds are going to be different.

srid:
> I removed the `srcSrc` call, but the tests still pass (on macOS at least).

Turns out this is because those `[[ .. == .. ]]` conditions always succeed on macOS's bash (3.2.57), but it is not a problem with bash from nixpkgs (5.2.15) - using which the tests fail as I expected.

Aside: the bash script is getting unwieldy; I'm starting to seriously think that we should write the tests is some other language.

srid:
I tested this PR in https://github.com/nammayatri/nammayatri/pull/109 and it does prevent rebuilds. Great!

nix/haskell-project.nix
srid: Thanks; added this in 600a179

srid (approved):
Gonna rebase and merge shortly.

nix/haskell-project.nix
```diff
@@ -55,14 +58,38 @@ in
                   hpack
                 ''
             else root;
+          makeSrcAutonomous = name: root: pkg: hlib.overrideSrc
+            {
+              src =
+                # Since 'root' maybe a subdirectory of another source
+                # derivation, the purpose of cleanSourceWith here is to create a
+                # new (smaller) source derivation that is a copy of 'root' but
+                # ignores the parent source contents.
```
roberth, line 67:
```suggestion
                # Since 'root' may be a subdirectory of a store path
                # (in string form, which means that it isn't automatically
                # copied), the purpose of cleanSourceWith here is to create a
                # new (smaller) store path that is a copy of 'root' but
                # does not contain the unrelated parent source contents.
```

It wasn't a derivation output at this point yet.

srid: Thanks; done.

Issue #36, opened 2022-11-01
Create app-target for each executable cabal package


HariAmoor-professional:
This ticket corresponds to @roberth's [comment](https://github.com/srid/haskell-flake/pull/34#discussion_r1010415656) on #34. To restate, we need to create a flake output in `apps'.*` for each executable cabal package. The logic to filter executable packages from non-executable ones is non-trivial.

srid:
@roberth @HariAmoor-professional Why do we need this? `nix run` already looks in `packages.*` if apps are not defined.

roberth:
I think it's relevant for multi-executable packages and packages whose executable has a different name than the package name.

srid:
How do we approach this?

- Proper way: parse the `.cabal` file, and extract executable stanzas.
- Quick-and-dirty way: look under `${drv}/bin` and use those executables.

roberth:
> parse the `.cabal` file

A regex on the cabal file contents may actually go a long way.
Isn't something like `^executable ([a-zA-Z0-9_.-]*)$` on any line a decent way to find all executables? Haven't checked the grammar, but this probably catches over 99% in practice.


> * look under `${drv}/bin`

We shouldn't do IFD to evaluate `attrNames apps`. That would slow down all apps evaluations, even if they're not from a haskell project.

srid:
Yea, we should parse the `.cabal` file. 

What should the interface be? In the context of #62, to begin with I think we could add an option under `outputs`, like `config.haskellProjects.default.outputs.localApps` that points to the apps themselves. Or do we want to expose them as per-package sets, like `outputs.localPackageInfo.<package-name>.<exe-name> = <nix-app>`?

roberth:
> EDIT: Or we can put this in the `passthru` of packages?


Putting it in `passthru` merges it with an IFD-based attrset, incurring the IFD cost, so that's not good enough for generating an apps attrset.

That shouldn't stop us from also adding information from passthru, although perhaps consistency with the rest of the ecosystem is a reason to make such improvements in cabal2nix instead.


> outputs.localPackageInfo.<package-name>.<exe-name> = <nix-app>

[...]

PR #137, opened 2023-04-04
WIP: POC implementation of localApps

sbh69840:
- [x] POC
- [ ] Move parsing to `find-haskell-paths` parser, add tests
- [ ] Use finalised options API (cf. ongoing discussion in comments)

POC for #36

### DIFF
--- a/nix/flake-module.nix
+++ b/nix/flake-module.nix
@@ -129,6 +129,14 @@ in
                   packages excluding everything else.
                 '';
               };
+              localApps = mkOption {
+                # TODO: use a more specific type
+                type = types.attrsOf types.raw;
+                readOnly = true;
+                description = ''
+                  The executable Haskell packages in the project.
+                '';
+              };
               devShell = mkOption {
                 type = types.package;
                 readOnly = true;
@@ -284,21 +292,19 @@ in
             let
               # Like mapAttrs, but merges the values (also attrsets) of the resulting attrset.
               mergeMapAttrs = f: attrs: lib.mkMerge (lib.mapAttrsToList f attrs);
+              mapKeys = f: key: attrs: lib.mapAttrs' (n: v: { name = f key n; value = v; }) attrs;
+              # Prefix package names with the project name (unless
+              # project is named `default`)
+              dropDefaultPrefix = name: packageName:
+                if name == "default"
+                then packageName
+                else "${name}-${packageName}";
             in
             {
               packages =
                 mergeMapAttrs
                   (name: project:
-                    let
-                      mapKeys = f: attrs: lib.mapAttrs' (n: v: { name = f n; value = v; }) attrs;
-                      # Prefix package names with the project name (unless
-                      # project is named `default`)
-                      dropDefaultPrefix = packageName:
-                        if name == "default"
-                        then packageName
-                        else "${name}-${packageName}";
-                    in
-                    lib.optionalAttrs project.autoWire (mapKeys dropDefaultPrefix project.outputs.localPackages))
+                    lib.optionalAttrs project.autoWire (mapKeys dropDefaultPrefix name project.outputs.localPackages))
                   config.haskellProjects;
               devShells =
                 mergeMapAttrs
@@ -313,6 +319,11 @@ in
                     lib.optionalAttrs (project.autoWire && project.devShell.enable && project.devShell.hlsCheck.enable) {
                       "${name}-hls" = project.outputs.hlsCheck;
                     })
+                  config.haskellProjects;
+              apps =
+                mergeMapAttrs
+                  (name: project:
+                    lib.optionalAttrs project.autoWire (mapKeys dropDefaultPrefix name project.outputs.localApps))
                   config.haskellProjects;
             };
         });
--- a/nix/haskell-project.nix
+++ b/nix/haskell-project.nix
@@ -38,6 +38,15 @@ in
         in
         lib.mapAttrs build-haskell-package config.packages;
 
+      # TODO: move this to `parser.nix` and add tests
+      parseExecutables = pkg:
+        let
+          cabalContents = builtins.readFile (lib.concatStringsSep "/" [ pkg.root (builtins.head (lib.filter (lib.strings.hasSuffix ".cabal") (builtins.attrNames (builtins.readDir pkg.root)))) ]);
+          regexExecutables = builtins.map (x: builtins.match ''^executable ([a-zA-Z0-9_.-]*)$'' x) (lib.strings.splitString "\n" cabalContents);
+          filterExecutables = builtins.map (x: builtins.head x) (builtins.filter (x: x != null) regexExecutables);
+        in
+        filterExecutables;
+
       defaultBuildTools = hp: with hp; {
         inherit
           cabal-install
@@ -84,6 +93,22 @@ in
           (name: _: finalPackages."${name}")
           config.packages;
 
+        localApps = builtins.foldl' (x: y: x // y) { }
+          (lib.mapAttrsToList
+            (name: pkg:
+              builtins.foldl' (x: y: x // y) { }
+                (builtins.map
+                  (x: {
+                    "${x}" = {
+                      type = "app";
+                      program = "${finalPackages.${name}}/bin/${x}";
+                    };
+                  })
+                  (parseExecutables pkg)
+                )
+            )
+            config.packages
+          );
         hlsCheck = runCommandInSimulatedShell
           devShell
           self "${projectKey}-hls-check"


srid:
> Use finalised options API

@sbh69840 For the API, let's go with this:

- Rename  the current`outputs.localPackages.<name>` to `outputs.packages.<name>.package`
- Rename this PR's `localApps.<exe-name>` to `outputs.packages.<name>.exes.<exe-name>`.

nix/haskell-project.nix
```diff
@@ -62,6 +71,19 @@ in
...
           };
       });
+      exes =
+        (name: value:
+          builtins.foldl' (x: y: x // y) { }
+            (builtins.map
+              (x: {
+                "${x}" = {
+                  type = "app";
+                  program = "${finalPackages.${name}}/bin/${x}";
```
srid, line 81: For the same reason as https://haskell.flake.page/docker#size

```suggestion
                  program = "${pkgs.haskell.lib.justStaticExecutables(finalPackages.${name})}/bin/${x}";
```

nix/find-haskell-paths/parser.nix
```diff
@@ -44,4 +44,44 @@ in
...
+                              c == "e" &&
+                              d == "c" &&
+                              e == "u" &&
+                              f == "t" &&
+                              g == "a" &&
+                              h == "b" &&
+                              i == "l" &&
+                              j == "e"
+                            then fail else (pure null)
+                          ))))))))));
```
srid, line 74: What is the purpose of this verbose `parseExec`? Why not use `parsec.string "executable"`?

srid: You could probably obviate thus using [`skipTill1`](https://github.com/kanwren/nix-parsec/blob/1bf25dd9c5de1257a1c67de3c81c96d05e8beb5e/parsec.nix#L566). @ursi might have better ideas.

ursi: I know of know reason to use this over `parsec.string`. Regarding anything fancier than than, I'd have to brush up on the library. In general my takeaway from the purescript parser was to use `parsec.regex` whenever possible for anything more complicatex than exact matches. If was much more performant.

sbh69840: Have pushed the change that uses `parsec.string`

nix/find-haskell-paths/parser.nix
```diff
@@ -44,4 +44,20 @@ in
         val;
     in
     parsec.runParser parser packageYamlFile;
+
+  # Extract all the executables from a .cabal file 
+  parseCabalExecutableName = cabalFile:
```
srid, line 49: Since this returns a list, `parseCabalExecutableNames` (i.e, plural)

nix/find-haskell-paths/parser_tests.nix
```diff
@@ -32,6 +32,23 @@ let
...
+    in
+    {
+      testSimple = {
+        expr = eval ''
+          executable foo-exec
+            main-is: foo.hs
+          executable bar-exec
+            main-is: bar.hs
+        '';
+        expected = [ "foo-exec" "bar-exec" ];
```
srid, line 49: Could expand the `expr` to insert lines before/between/after these executable stanzas?

nix/flake-module.nix
```diff
@@ -125,14 +125,15 @@ in
                   overrides, on top of `basePackages`.
                 '';
               };
-              localPackages = mkOption {
-                type = types.attrsOf types.package;
+              packages = mkOption {
+                # TODO: use a more specific type
+                type = types.attrsOf types.raw;
```
srid, line 130: `types.submodule`?

nix/haskell-project.nix
```diff
@@ -38,6 +38,19 @@ in
         in
         lib.mapAttrs build-haskell-package config.packages;
 
+      # TODO: move this to `parser.nix` and add tests
+      parseExecutables = pkg:
+        let
```
srid, line 43: - Rename `find-haskell-paths/` to `nix-haskell-parsers`
- Have `nix-haskell-parsers/default.nix` return a attrset of two functions, including this one.

nix/haskell-project.nix
```diff
@@ -62,6 +75,19 @@ in
               ++ builtins.attrValues (config.devShell.extraLibraries p);
           };
       });
+      exes =
+        (name: value:
+          builtins.foldl' (x: y: x // y) { }
+            (builtins.map
+              (x: {
+                "${x}" = {
+                  type = "app";
```
srid, line 84: I think the `type` here is redundant. `"${x}".program = ...` should work AFAIU flake-parts.

nix/haskell-project.nix
```diff
@@ -62,6 +75,19 @@ in
               ++ builtins.attrValues (config.devShell.extraLibraries p);
           };
       });
+      exes =
+        (name: value:
```
srid, line 79: This should be refactored.

srid (changes requested):

nix/flake-module.nix
```diff
@@ -291,21 +292,21 @@ in
             let
               # Like mapAttrs, but merges the values (also attrsets) of the resulting attrset.
               mergeMapAttrs = f: attrs: lib.mkMerge (lib.mapAttrsToList f attrs);
+              # Prefix package names with the project name (unless
+              # project is named `default`)
+              dropDefaultPrefix = name: packageName:
+                if name == "default"
+                then packageName
+                else "${name}-${packageName}";
```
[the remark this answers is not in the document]
sbh69840: This function is called in `mapKeys` and the parameter passed is the `name` of the project.
https://github.com/srid/haskell-flake/blob/d82e2ccdf36e4fc0e1576e2f4b1f7a216fb390c1/nix/flake-module.nix#L307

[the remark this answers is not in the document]
srid: Huh, you're right. I wrote `mapKeys` wrong. Its `f` should've been an unary function such as to be invoked as `mapKeys (dropDefaultPrefix name) ...`.

srid:
Just merged from `master` (because, using this branch in nammayatri)

nix/haskell-project.nix
sbh69840: I have named it haskell-parsers as it is already under the directory nix

nix/flake-module.nix
sbh69840: `outputs.package` is an attrset where attribute keys are `<packageName>`. The closest possible type I see is `types.attrsOf types.attrs`. What do you think?

nix/flake-module.nix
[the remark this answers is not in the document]
sbh69840: I will make this change!

nix/flake-module.nix
srid: @sbh69840 It should be like this:

https://github.com/srid/haskell-flake/blob/8fb3bbe77c74f2b7249d769d2eb462df917bbdc1/nix/flake-module.nix#L290

which is then defined as:

https://github.com/srid/haskell-flake/blob/8fb3bbe77c74f2b7249d769d2eb462df917bbdc1/nix/flake-module.nix#L155-L161

So you are looking for something like:

```nix
let 
  packageInfoSubmodule = types.submoduleWith {
    modules = [
      ({ config, name, ...}: {
        # 'name' refers to package name
        options = ...
      })
    ];
  };
in
  options.outputs.packages = mkOption { type = packageInfoSubmodule; ...
```

srid:
@sbh69840 Could you also modify the test https://github.com/srid/haskell-flake/blob/master/test/simple/test.sh to check that https://github.com/srid/haskell-flake/blob/8fb3bbe77c74f2b7249d769d2eb462df917bbdc1/test/simple/haskell-flake-test.cabal#L12 exists.

Something like `nix eval .#apps.<system>.<exe>.program` I think.

srid (changes requested):
@sbh69840 Sorry, I've completely overlooked it, but just caught this issue.

The PR in its current form is evaluating `find-haskell-packages/default.nix` in **two** places -- flake-module.nix and haskell-project.nix.

image

That's wrong; here's why:

In a multi-package project, remove the `cabal.project` file and explicitly specify the `packages` option. Run the flake app, and you'll see an evaluation error. 

For example, clone this branch https://github.com/srid/haskell-multi-nix/pull/5 and run

```
❯ nix run           
warning: Git tree '/Users/srid/code/haskell-multi-nix' is dirty
error: Neither a .cabal file nor a package.yaml found under /nix/store/5avkn71wd7s20kzslyr0hympg594mhhm-source
(use '--show-trace' to show detailed location information)
```

---

These parsers are designed to fail, and meant to be used in `options` rather than `config` which explains the problem. So how do we solve this? One approach is to allow users to explicitly specify `executables` just as they already do for `root`, a la:

```nix
haskellProjects.default = {
  packages = {
    foo = {
      root = ./foo;
      executables = [ "foo" ];
    };
  };
}
```

Then, we can use `find-haskell-packages` to determine the default value for `packages`, without needing to run the parsers at the `config` level.

srid:
However, what if the user specifies:

```nix
haskellProjects.default = {
  packages = {
    foo = {
      root = ./foo;
    };
  };
}
```

ie., `root` specified but no `executables`. I think we should be able to find the default value for it by parsing the .cabal file during `options` default. A test should be added to reflect this case. 

More generally, we are decoupling the metadata extraction (from .cabal, etc.) from where they get used (haskell-project.nix).

sbh69840:
In the initial POC commit I was parsing the cabal file based on the `root` attribute from `config.packages.<packageName>`. This would work in cases where `cabal.project` is present or not.

srid:
> In the initial POC commit I was parsing the cabal file based on the `root` attribute from `config.packages.<packageName>`. This would work in cases where `cabal.project` is present or not.

That's simpler; let's go with that (assuming the parser is kept in the sub-directory). You could make `default.nix` return an attrset of two functions (instead of single function).

srid:
@roberth Hmm, this commit https://github.com/srid/haskell-flake/pull/137/commits/3ee7c78ad830dbd50ac537ea2d8f09aa0a83000d (I simply moved the code around) breaks the documentation test

https://github.com/srid/haskell-flake/actions/runs/4746595660/jobs/8430433835

```
 error: Most nixpkgs attributes are not supported when generating documentation.
       Please check with `--show-trace` to see which option leads to this `pkgs.callPackage` reference. Often it can be cut short with a `defaultText` argument to `lib.mkOption`, or by escaping an option `example` using `lib.literalExpression`.
```

srid (approved):

PR #143, opened 2023-04-12
Change `autoWire` to be an enum type

### LINKED ISSUE #142: Granular autowiring
In #134 we added an `autoWire` option. Setting this to false disables wiring of **all** outputs. But often, we want to turn off autowiring of just a particular output; when using #135, this would be to turn off autowiring of devShells but keep autowiring of everything else (packages & checks), which helps avoid having to write verbose manual wiring of say `packages` as seen here:

https://github.com/nammayatri/beckn-gateway/blob/2625936af5273bd024cb09d99072077485f33b22/flake.nix#L26-L33

(This example has just two packages but there are repos with >10 local packages).

Proposal: replace the current `autoWire: bool` with:

```nix
haskellProject.default = {
  autoWire = {
    devShells = false;
    packages = true;
    checks = true;
  };
}
```

We could also support `autoWire = null` (using `types.nullOr`) as a special case to turn off all autowiring (mirroring the existing `autoWire = false` behaviour).

srid:
Resolves #142 

(Not removing existing options in this PR)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - #134: Add `autoWire` option to control generation of flake outputs
     - #138: Add `checks` to `outputs` submodule
-    - #??: Changed `autoWire` to be an enum type, for granular controlling of which outputs to autowire.
+    - #143: Changed `autoWire` to be an enum type, for granular controlling of which outputs to autowire.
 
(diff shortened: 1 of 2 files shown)


srid:
To disable just the devShell autowiring:

image

Issue #101, opened 2023-03-01
A simple interface to `overrides`


srid:
Just an umbrella issue tracking our discussions around this topic.

See https://github.com/srid/haskell-flake/pull/87#issuecomment-1450286992 for a summary on our initial exploration.

srid:
Currently, we write them as follows:

```nix
{
  source-overrides = {
    foo = inputs.fooRepo + /foo;
    qux = "1.2.3"; # Hackage version; yea I just learned that this is already possible.
  };
  overrides = self: super: with pkgs.haskell.lib.compose; {
    qux = lib.pipe super.qux [ dontCheck dontHaddock ];
  };
}
```

This can be made nicer using:

```nix
let
  # A function that enables us to write `foo = [ dontCheck ]` instead of `foo =
  # lib.pipe super.foo [ dontCheck ]` in haskell-flake's `overrides`.
  compilePipe = f: self: super:
    lib.mapAttrs
      (name: value:
        if lib.isList value then
          lib.pipe super.${name} value
        else
          value
      )
      (f self super);
in 
{
  source-overrides = {
    foo = inputs.fooRepo + /foo;
    qux = "1.2.3"; # Hackage version; yea I just learned that this is already possible.
  };
  overrides = self: super: with pkgs.haskell.lib.compose; {
    qux = [ dontCheck dontHaddock ];
  };
}
```

It simply avoids having to specify `lib.pipe super.qux`. This can make visible difference when you have lots of overrides; see https://github.com/nammayatri/nammayatri/pull/47/commits/1374db49425e41698630e419b551b60461e26903 of https://github.com/nammayatri/nammayatri/pull/47. Not sure if worth upstreaming to `haskell-flake` in some form ...

srid:
Here's an idea, and it would also solve the problem of "allowing users to specify default overrides for _all_ local packages _at once_" (to obviate having #146 in haskell-flake).

```nix
{
  overrides = {
    # Overrides for all local packages
    localDefaults = info: [
      dontHaddock
      disableLibraryProfiling
    ] ++ lib.optionals (info.exes != { }) [
      enableSeparateBinOutput
    ];
    # Usual overrides on a per package basis, any package.
    perPackage = _self: _super: {
      aeson = [ dontCheck ];
    };
  };
}
```

[...]

PR #146, opened 2023-04-25
Use `enableSeparateBinOutput` to prevent duplicate builds

srid:
Aims to resolve https://github.com/nammayatri/nammayatri/issues/687

Basically, the use of `justStaticExecutables` causes the library/executable to be built **twice**!

---

Normal library build:

```
❯ nix build github:srid/devour-flake --no-link --print-out-paths --override-input flake github:srid/haskell-template/bdd135025466fc82fbc88a46043c53ec28930979 2> /dev/null | xargs cat | tr ' ' '\n' | grep -E "\-haskell-template-0.1.0.0$" | uniq
/nix/store/kv6yrjqs08nxirqvwjbi7lm763lbizcz-haskell-template-0.1.0.0
```

With [`justStaticExecutables`](https://github.com/srid/haskell-template/commit/0bbc2493d0891cb57cc2a926466bfb562557019c),

```
❯ nix build github:srid/devour-flake --no-link --print-out-paths --override-input flake github:srid/haskell-template/0bbc2493d0891cb57cc2a926466bfb562557019c 2> /dev/null | xargs cat | tr ' ' '\n' | grep -E "\-haskell-template-0.1.0.0$" | uniq
/nix/store/ra02zpjymb3fyjsbaln661k13ayrnyws-haskell-template-0.1.0.0
/nix/store/di3798v15v4q6db10c6sx87664gxrfzp-haskell-template-0.1.0.0
```

---

The solution here seems to be to use `enableSeparateBinOutput`. This PR in the current form enables this flag for **all** packages. Not sure if that's desirable; so we may want to have an option to turn it off (`packages.<name>.flags.separateBinOutput`?).

### DIFF
--- a/nix/build-haskell-package.nix
+++ b/nix/build-haskell-package.nix
@@ -61,4 +61,9 @@ lib.pipe pkg
     # Make sure all files we use are included in the sdist, as a check
     # for release-worthiness.
     fromSdist
+
+    # TODO: Make it an option that the user can override
+    # This is better than using justStaticExecutables, because with the later
+    # builds will repeated twice!
+    pkgs.haskell.lib.enableSeparateBinOutput
   ]
--- a/nix/haskell-project.nix
+++ b/nix/haskell-project.nix
@@ -68,16 +68,13 @@ in
         package = finalPackages.${name};
         exes =
           let
-            inherit (pkgs.haskell.lib) enableSeparateBinOutput;
             exeNames = haskell-parsers.getCabalExecutables value.root;
-            # TODO: We must do this globally in overlay, to avoid repeat builds?
-            packageBinOut = (enableSeparateBinOutput finalPackages.${name}).bin;
           in
           lib.listToAttrs
             (map
               (exe:
                 lib.nameValuePair exe {
-                  program = "${packageBinOut}/bin/${exe}";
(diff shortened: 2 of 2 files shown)


srid:
> This PR in the current form enables this flag for **all** packages. Not sure if that's desirable; so we may want to have an option to turn it off (`packages.<name>.flags.separateBinOutput`?).

Might as well use this opportunity to enable other flags (like `dontHaddock`) on local packages en masse.

srid:
> ... or add an option?

Consider this in the context of  #101 ?

srid:
I get

```
error: cycle detected in build of '/nix/store/xx1j8g8ikrfi0s3gxiz052an73sqzlmj-emanote-1.0.3.5.drv' in the references of output 'bin' from output 'out'
``` 

in emanote, https://github.com/srid/emanote/pull/435

srid:
I don't think haskell-flake should do this. Instead, let the user do it after providing them with a way to specify default overrides for local packages, per https://github.com/srid/haskell-flake/issues/101#issuecomment-1526276493

srid:
Closing in favour of https://github.com/srid/haskell-flake/pull/162

srid:
Doing this in https://github.com/srid/haskell-flake/pull/162/commits/421d90674e7d04c58ef6bf5b82832a5a807e5e01

PR #149, opened 2023-04-26
Run `cleanSourceWith` *before* `cabal2nix`

srid:
Also make haskell-flake produce more debug output (if `debug` is enabled by the user)


srid:
I've rebased this on top of #150 to show that this PR resolves #145 (assuming that the comparison, in #150, of the `cabal2nixDeriver` hash alone is sufficient to test this).

PR #175, opened 2023-06-23
`benchmark` and `hoogle` option in devShellSubmodule

shivaraj-bh:
The `shellFor` function allows specifying the `doBenchmark` attribute to include the benchmark's `build-depends` from the `.cabal` file in the derivation. We have to expose this option as an attribute in the `devShellSubmodule` of haskell-flake.
 
As a hack this attribute input can still be passed to `shellFor` like, `devShell.mkShellArgs.doBenchmark = true;` but that's not the intended behaviour of `mkShellArgs` option.

In addition to the above change, this PR also makes `withHoogle` input to `shellFor` also a configurable option under `devShellSubmodule`.

### DIFF
--- a/nix/modules/project/devshell.nix
+++ b/nix/modules/project/devshell.nix
@@ -70,6 +70,7 @@ let
 
       hoogle = mkOption {
         type = types.bool;
+        default = true;
         description = ''
           Whether to include Hoogle in the development shell.
           The value of this option will set the corresponding `withHoogle` flag in the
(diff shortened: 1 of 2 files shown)


nix/modules/project/devshell.nix
```diff
@@ -100,6 +109,7 @@ in
             (name: p."${name}")
             (lib.attrNames localPackages);
         withHoogle = true;
```
srid, line 111: Can we make this overridable as well (in this or separate PR)?

srid (approved):

PR #195, opened 2023-08-27
Extend README.md to focus on the benefits

malteneuss:
I felt that that the Readme was not conveying how convenient this project is for Haskell development in Nix so i extended it.

@srid Feel free to edit, request changes or delete this PR.
And thanks for this great tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,16 +2,9 @@
 
 <img src="./doc/logo.webp" width=100 />
 
-There are several ways to manage Haskell packages using Nix with varying degrees of integration.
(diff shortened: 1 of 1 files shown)


srid:
This is great, thanks! I should also use this in https://zero-to-flakes.com/haskell-flake

malteneuss:
Sure. Feel free to copy over as much as you like. I would like to see something like this in the Github project Readme to present the benefits to the user and a feel on how to use this software on the first point of contact (i think github page is ranked much higher than the zero-to-flakes page)

PR #203, opened 2023-11-07
`devShell.extraLibraries`: abort when using an older commit of nixpkgs

shivaraj-bh:
`extraDependencies` was added as an argument to `shellFor` only after: https://github.com/NixOS/nixpkgs/commit/2e87d165f74411ae00f964a508945696969ff53d

### DIFF
--- a/nix/modules/project/devshell.nix
+++ b/nix/modules/project/devshell.nix
@@ -28,7 +28,7 @@ let
         default = hp: { };
       };
       extraLibraries = mkOption {
-        type = functionTo (types.attrsOf (types.nullOr types.package));
+        type = types.nullOr (functionTo (types.attrsOf (types.nullOr types.package)));
         description = ''
           Extra Haskell libraries available in the shell's environment.
           These can be used in the shell's `runghc` and `ghci` for instance.
@@ -37,7 +37,9 @@ let
 
           The return type is an attribute set for overridability and syntax, as only the values are used.
         '';
-        default = hp: { };
+        default = null;
+        apply = f: hp:
+          if f == null then { } else f hp;
         defaultText = lib.literalExpression "hp: { }";
         example = lib.literalExpression "hp: { inherit (hp) releaser; }";
       };
@@ -130,7 +132,10 @@ in
                 ++ builtins.attrValues (config.devShell.extraLibraries p);
             }
           else
-            config.log.traceWarning "Your nixpkgs does not support `extraDependencies` in `shellFor`." null;
+            if config.devShell.extraLibraries != null then
+              builtins.throw "The 'extraLibraries' option is only available when using a version of nixpkgs that supports extraDependencies in shellFor."
+            else
+              null;
       });
 
     in
--- a/nix/modules/project/devshell.nix
+++ b/nix/modules/project/devshell.nix
@@ -40,7 +40,7 @@ let
         default = null;
         apply = f: hp:
           if f == null then { } else f hp;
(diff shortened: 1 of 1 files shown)


nix/modules/project/devshell.nix
```diff
@@ -121,12 +121,16 @@ in
...
+        extraDependencies =
+          if lib.hasAttr "extraDependencies" (lib.functionArgs finalPackages.shellFor) then
+            p:
+              let o = mkShellArgs.extraDependencies or (_: { }) p;
+              in o // {
+                libraryHaskellDepends = o.libraryHaskellDepends or [ ]
+                  ++ builtins.attrValues (config.devShell.extraLibraries p);
+              }
+          else
+            config.log.traceWarning "Your nixpkgs does not support `extraDependencies` in `shellFor`." null;
```
srid, line 133: I think we should also fail the build right away (using `throw`, possibly) as well.

We don't want users to specify a `extraDependencies` and then wonder why they are not available in the dev shell.

srid: Also, perhaps we can do this logging/throw'ing in the `mkOption` definition itself?

See https://github.com/srid/haskell-flake/blob/4f6a2cbe4ee9024ffebb834473592db6c6b2c943/nix/modules/project/defaults.nix#L57

shivaraj-bh: I was considering to throw an error based on whether the user has set the option `config.devShell.extraLibraries` or not. Is there an easy way I could detect if this option, when given a package set as argument, returns an empty attrset? I am asking this because in the `else` block I won't have the argument `p` to evaluate `config.devShell.extraLibraries`.

shivaraj-bh: I can't do this:
```
extraDependencies = p:
if ...  then
...
else
...
```
because then `extraDependencies` is not null when `extraLibraries` returns an empty set, it is a function -- and that would cause an error: `Cannot coerce a function to a string`

srid: Something like this? (Haven't tested it)

```diff
diff --git a/nix/modules/project/devshell.nix b/nix/modules/project/devshell.nix
index c8867f7..0317333 100644
--- a/nix/modules/project/devshell.nix
+++ b/nix/modules/project/devshell.nix
@@ -28,7 +28,7 @@ let
         default = hp: { };
       };
       extraLibraries = mkOption {
-        type = functionTo (types.attrsOf (types.nullOr types.package));
+        type = types.nullOr (functionTo (types.attrsOf (types.nullOr types.package)));
         description = ''
           Extra Haskell libraries available in the shell's environment.
           These can be used in the shell's `runghc` and `ghci` for instance.
@@ -37,9 +37,14 @@ let
 
           The return type is an attribute set for overridability and syntax, as only the values are used.
         '';
-        default = hp: { };
+        default = null;
         defaultText = lib.literalExpression "hp: { }";
         example = lib.literalExpression "hp: { inherit (hp) releaser; }";
+        apply = f: hp: 
+            if f == null then { }
+            else if lib.hasAttr "extraDependencies" (lib.functionArgs hp.shellFor)
+              then f hp
+              else builtins.throw "The 'extraLibraries' option is only available when using a version of nixpkgs that supports extraDependencies in shellFor.";
       };
 
       mkShellArgs = mkOption {

```

shivaraj-bh: Thanks! I have taken some bits from here and updated the PR

PR #204, opened 2023-11-11
modified callPackage based on custom-default.nix file.

aravindgopall:


srid (commented):
[As briefly talked with @aravindgopall today]

PR #204 takes the approach of externalizing the generated Nix expr to the respective repo, but this won't work for other dependency overrides like those that come from Hackage.

Since both `callCabal2nix` and `callHackage` use `haskellSrc2nix` underneath (which is then [`callPackage`'ed](https://github.com/NixOS/nixpkgs/blob/77e3fa92e8a6a36c3b97fb574b06de7fc7a747c8/pkgs/development/haskell-modules/make-package-set.nix#L179)) - we can decouple the exact point where `haskellSrc2nix` is run. Basically cache the output of `haskellSrc2nix` in a set of local files (`./.haskellSrc2nix/aeson-2.0.0.0.nix`, etc.) as part of a flake app. Then have haskell-flake use that cache instead of evaluating using `call*` family of functions (this behaviour can be toggled using a flag). Add a flake check and/or a pre-commit hook to check the consistency of this cache. 

This way, we should be able to avoid IFD without needing to define our own package set from scratch. ie., we still use `pkgs.haskellPackages` but generate only the local and overridden dependencies.

I'll take a look at exploring this.

EDIT: Discussion on this at https://nixos.zulipchat.com/#narrow/stream/413949-haskell-flake/topic/Removing.20IFD

srid:
Superceded by #231

PR #220, opened 2024-01-31
Allow user to disable use of `buildFromCabalSdist`

srid:
Using `buildFromCabalSdist`, as we do here:

https://github.com/srid/haskell-flake/blob/738417211e16514d24d0e247bb83781e12293342/nix/build-haskell-package.nix#L15-L16

https://github.com/srid/haskell-flake/blob/738417211e16514d24d0e247bb83781e12293342/nix/build-haskell-package.nix#L40-L43

is problematic, because it uses its own `cabal-install` from a **different** package set called [`buildHaskellPackages`](https://github.com/NixOS/nixpkgs/blob/c33213823590ca56d8c7aadb125dd62ab3d139ac/pkgs/development/haskell-modules/make-package-set.nix#L9C3-L9C23). Thus, the user's own own overrides are not respected when building `cabal-install`.

See:

1. https://nixos.zulipchat.com/#narrow/stream/413949-haskell-flake/topic/cabal-install.20ignoring.20user's.20package.20set
2. https://nixos.zulipchat.com/#narrow/stream/413949-haskell-flake/topic/Using.20horizon.20package.20set.20with.20haskell-flake (where it prevents use of horizon package sets from @locallycompact)

---

To make this user configurable I made it part of the package settings (`all.nix`). So toggling this is now same as toggling haddock/jailbreak/check and the like, viz. this patch for https://github.com/srid/haskell-multi-nix :

```diff
--- a/flake.nix
+++ b/flake.nix
@@ -13,6 +14,7 @@
           projectFlakeName = "haskell-multi-nix";
           # Want to override dependencies?
           # See https://haskell.flake.page/dependency
+          settings.foo.buildFromCabalSdist = false;
         };
         packages.default = self'.packages.bar;
       };
```

### DIFF
--- a/nix/build-haskell-package.nix
+++ b/nix/build-haskell-package.nix
@@ -12,9 +12,6 @@
 }:
 
 let
-  fromSdist = self.buildFromCabalSdist or
-    (log.traceWarning "Your nixpkgs does not have hs.buildFromCabalSdist" (pkg: pkg));
-
   mkNewStorePath = name: src:
     # Since 'src' may be a subdirectory of a store path
     # (in string form, which means that it isn't automatically
@@ -36,9 +33,4 @@ lib.pipe root
 
     (root: self.callCabal2nix name root { })
     (x: log.traceDebug "${name}.cabal2nixDeriver ${x.cabal2nixDeriver.outPath}" x)
-
-    # Make sure all files we use are included in the sdist, as a check
-    # for release-worthiness.
-    fromSdist
-    (x: log.traceDebug "${name}.fromSdist ${x.outPath}" x)
   ]
--- a/nix/modules/project/defaults.nix
+++ b/nix/modules/project/defaults.nix
@@ -89,6 +89,7 @@ in
               # Disabling haddock and profiling is mainly to speed up Nix builds.
               haddock = lib.mkDefault false; # Because, this is end-user software. No need for library docs.
               libraryProfiling = lib.mkDefault false; # Avoid double-compilation.
+              buildFromCabalSdist = lib.mkDefault true;
             };
         in
         if config.defaults.enable then localSettings else { };
--- a/nix/modules/project/packages/default.nix
+++ b/nix/modules/project/packages/default.nix
@@ -62,7 +62,7 @@ in
             if lib.types.path.check cfg.source
             then
               log.traceDebug "${name}.callCabal2nix ${cfg.source}"
-                (build-haskell-package name cfg.source)
+                (cfg.transform (build-haskell-package name cfg.source))
             else
               log.traceDebug "${name}.callHackage ${cfg.source}"
                 (self.callHackage name cfg.source { });
--- a/nix/modules/project/packages/default.nix
+++ b/nix/modules/project/packages/default.nix
@@ -62,7 +62,7 @@ in
             if lib.types.path.check cfg.source
             then
               log.traceDebug "${name}.callCabal2nix ${cfg.source}"
-                (cfg.transform (build-haskell-package name cfg.source))
+                (build-haskell-package name cfg.source)
             else
               log.traceDebug "${name}.callHackage ${cfg.source}"
                 (self.callHackage name cfg.source { });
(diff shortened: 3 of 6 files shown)


srid:
@roberth What's your opinion on leaving this option disabled by _default_ (as opposed to enabling it by default)?

Because [these kinds of issues](https://nixos.zulipchat.com/#narrow/stream/413949-haskell-flake/topic/cabal-install.20ignoring.20user's.20package.20set) are not straightforward to debug without knowledge of the nixpkgs Haskell infrastructure.

roberth:
We should try not to add more configuration for things that should just work, and haskell-flake has a role in CI to ensure release-worthiness. Building from sdist is an important part of that.

Either using `buildHaskellPackages` was a mistake (not sure; I believe it makes cross-compilation work), or we should apply the user overrides to `buildHaskellPackages` too. That's how Nixpkgs overlays work. It has a special `crossOverlays` in case you want to edit the final package set only, whereas `overlays` applies to all instances of the Nixpkgs fixpoint.

Taking a step back though, what's the point of having a broken `buildFromSdist` in a haskell package set when it's cross? Arguably that'd be a bug in the Nixpkgs Haskell infrastructure. What if we change it to just take that function from the `basePackages`? Does that really break anything? If it does, it can and should be fixed in Nixpkgs and/or Horizon. (Probably by making it use `buildHaskellPackages.cabal-install` internally, and managing that package set correctly, if needed.)

> not straightforward to debug without knowledge of the nixpkgs Haskell infrastructure.

I see this as the purpose of tools like `haskell-flake`: to distill all such knowledge into code that just does the right thing.
A corollary to this is that we risk encoding ignorance instead, and making its true resolution harder to accomplish.

srid:
> What if we change it to just take that function from the `basePackages`? Does that really break anything?

I tried using `basePackages.buildFromCabalSdist` (https://github.com/srid/haskell-flake/compare/transform2) but that didn't do anything different.

Here's the repro for the (1) case in the PR description:

https://github.com/srid/rings/commit/f8864c36ec19a8185fa070d1b4480b0d5904b720

Since I can't contribute to nixpkgs anyway, I'd be in favour of keeping this option (albeit as undocumented), but your points are all valid - so we should leave it enabled by default. Obviously if someone resolves this upstream we wouldn't even need this PR. Is there an official nixpkgs bug for this?

roberth:
> Is there an official nixpkgs bug for this?

Doesn't seem so; no open issues for the `buildFromSdist` or `buildFromCabalSdist` functions.


> I tried using `basePackages.buildFromCabalSdist` (https://github.com/srid/haskell-flake/compare/transform2) but that didn't do anything different.

Shouldn't that use the right `cabal-install`? Maybe it's not just about `Cabal`/`cabal-install`,` but caused by a custom `Setup.hs`?
Did you try `buildFromSdist`?

Note the docs on the `cabal` variations in `pkgs/development/haskell-modules/make-package-set.nix`

      Run `cabal sdist` on a source.

      Unlike `haskell.lib.sdistTarball`, this does not require any dependencies
      to be present, as it uses `cabal-install` instead of building `Setup.hs`.
      This makes `cabalSdist` faster than `sdistTarball`.

If `buildFromSdist` solves the problem, I think we can
- Use `buildFromSdist` when a `Setup.hs` is present, and/or it has a component with a `build-type:` that isn't `Simple`
- In nixpkgs, maybe speed up `buildFromSdist` by only forwarding the `setupHaskellDepends` - not the `libraryHaskellDepends` etc. Not sure if `./Setup.hs sdist` works without library deps, but IMO it _should_.
   - If this doesn't work out, we only lose a bit of build latency, because of less concurrency, but no extra work is performed, fwiw. Also only in cases where `./Setup.hs` exists.

> Since I can't contribute to nixpkgs anyway

I could help out with the communication if needed.

roberth:
Or perhaps even ignore the little performance optimization and just go for `buildFromSdist`.

srid:
> Or perhaps even ignore the little performance optimization and just go for `buildFromSdist`.

Let's do that then: https://github.com/srid/haskell-flake/pull/221

PR #221, opened 2024-01-31
Switch to `buildFromSdist`

srid:
Closes #220 

`buildFromCabalSdist` has problems; see #220 

So we are switching to the slightly less performant `buildFromSdist`.


srid:
@roberth I assume you are okay with this. I'll merge this shortly.

roberth (approved):

PR #222, opened 2024-02-05
fix(`cabal.project` parser): parse the packages ending with eof

shivaraj-bh:
To reproduce:
```sh
git clone git@github.com:srid/haskell-flake.git
curl https://raw.githubusercontent.com/shivaraj-bh/haskell-flake/c80a145a228631161a16fcbb5edbc6a5171803ad/test/with-subdir/cabal.project > test/with-subdir/cabal.project
, test
```


srid (changes requested):
Could you add a unit test for this case (ending with eof) here?

https://github.com/srid/haskell-flake/blob/c301f31092a74a6acde9d98d6431ea0a9952ca0f/nix/haskell-parsers/parser_tests.nix#L5-L20

shivaraj-bh:
@srid unit test added

PR #231, opened 2024-02-16
Cache cabal2nix expr to avoid IFD

### LINKED ISSUE #144: IFD-less evaluation of local packages
IFD can really slow down the evaluation of the flake, such as to make the launching of `nix develop` noticeably slow -- this is currently the case for https://github.com/nammayatri/nammayatri which has gazillion local packages with lots of dependencies.

The idea here is to do something similar to the `hpack -> cabal` workflow, but for `cabal -> nix`. ie., use `cabal2nix` to generate a `default.nix` in the local package directory ([example](https://github.com/NorfairKing/autodocodec/blob/master/autodocodec/default.nix)), for which we provide a `nix run`'able flake app that the developer will run anytime changing the .cabal file. There will also be a flake check (analogous to #97) to make sure that generated `default.nix` is in sync with the .cabal file (useful in CI).

We could actually name the `default.nix` using something unique to haskell-flake: `haskell-package.nix`.

Then, the following configuration:

```nix
{
  haskellProjects.default = {
    packages = {
      foo.root = ./foo;
    };
}
```

... would look for `haskell-packages.nix` and use that; if not, it looks for the .cabal file falling back to using IFD (`callCabal2nix`).

cc @roberth @locallycompact @NorfairKing for feedback/suggestions/ideas.

srid:
cc @aravindgopall who continues this.

Resolves #144


roberth:
A good CI will cache it for you.
If you want to avoid IFD, you could perhaps look for a local file, and rely on something like the `cabal2nix` [pre-commit hook](https://flake.parts/options/pre-commit-hooks-nix).

Issue #270, opened 2024-03-12
devShell to eagerly evaluated? 


MangoIV:
to reproduce 
- set `basePackages = pkgs.haskell.packages.ghc90`
- set `devShell.enable = false`

this should eval error: 
>  error: haskell-language-server has dropped support for ghc 9.0 in version 2.4.0.0, please use a newer ghc version or an older nixpkgs version

I'm suspecting that there is some issue with the `devShell` being evaluated to eagerly; the package shouldn't depend on `haskell-language-server` and doesn't try to evaluate/ build it in all other projects

srid:
Does that PR fix it? Because `devShell.tools` already uses `typs.lazyAttrsOf`.

https://github.com/srid/haskell-flake/blob/938888c7ec66d265dda2dad9f4f4ffc9a5a30a78/nix/modules/project/devshell.nix#L19-L20

roberth:
I've tried `hercules-ci-agent`, but got stuck migrating away from its haskell overlay
- https://github.com/srid/haskell-flake/pull/162#issuecomment-1992533135.

roberth:
> `devShell.tools` already uses `typs.lazyAttrsOf`.

According to the trace I got from the haskell-flake `example/`, the culprit was `basePackages`, which was fixed by the PR.
I don't know @MangoIV's project though.

PR #271, opened 2024-03-12
attrsOf -> lazyAttrsOf

roberth:
This means we don't support `mkIf` in the exact places where the type was `attrsOf`. Almost always, `mkIf` can be replaced by `optionalAttrs`.

The benefit is that the code becomes much lazier, improving performance and
- Fixes https://github.com/srid/haskell-flake/issues/270

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Breaking changes
   - #221: Switch to `buildFromSdist`, to allow using non-standard package sets (wherein `cabal-install` is otherwise built without using user's overrides)
   - #253: Turn off `buildFromSdist` by default. It can now be enabled manually by setting `settings.<name>.buildFromSdist` to `true`.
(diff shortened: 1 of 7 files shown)


srid:
Thanks! 

I guess I should just use `lazyAttrsOf` for everything by default? What's the point of keeping the strict version by default?

roberth:
In Nixpkgs, `attrsOf` is often "just data" which is consumed wholesale. There it *might* make sense to still use `attrsOf` and support `mkIf`, but I'm not completely convinced.
I think the rule "`mkIf` for submodules, `optionalAttrs` for attrsets" works at least as well as the status quo, but I haven't proven it, and everyone will need to be made aware of it, if Nixpkgs does switch.

roberth:
I've opened a Nixpkgs issue for this topic.
- https://github.com/NixOS/nixpkgs/issues/295419

PR #274, opened 2024-03-13
Add a test: avoid rebuilds for top-level package

srid:


srid:
https://github.com/srid/haskell-flake/pull/276

Issue #273, opened 2024-03-13
Add an `overlays` option


srid:
While people can use `basePackages = pkgs.haskellPackages.extend ...` to already do this, let's add an `overlays` option anyway for efficiency purposes (see https://github.com/srid/haskell-flake/pull/162#issuecomment-1992571391).

However, unlike `packages` and `settings` the `overlays` type should probably be made un-mergeable when an user imports a haskell-flake project module, since the order is not something the user can specify. 

Ultimately, most users would just use `packages` and `settings` - whereas the small subset who would want to stay close to overlays can still specify them, whilst not being able to use the haskell-flake shareable module system.

srid:
https://github.com/srid/haskell-flake/pull/277

PR #277, opened 2024-03-14
Add `otherOverlays` option

srid:
#273 

@roberth Could you try this out?


srid:
(CI won't be up, as I'm in a library, and have VMs turned off 🤓)

roberth:
Works for `hercules-ci-agent`.
It'd be good to have a test for this option.

Issue #285, opened 2024-03-29
Proposal: merge `packages` and `settings` submodules


srid:
Context: https://nixos.zulipchat.com/#narrow/stream/413949-haskell-flake/topic/.E2.9C.94.20.60custom.60.20setting.20stopped.20working/near/430086733

Instead of separate `packages.<name>.*` and `settings.<name>.*`, why not have a single submodule namespace? Isn't that simpler DX?

viz.: Instead of doing:

```nix
{
  packages = {
    aeson.source = "1.5.0";
    myhaskell.source = ./haskell;
  };
  settings = {
    hello_rust.custom =self'.packages.hello_rust;
    myhaskell.libraryProfiling = true;
  };
}
```

why not just:

```nix
{
  packages = {
    hello_rust.source = self'.packages.hello_rust
    myhaskell.source = ./haskell;
    myhaskell.setting.libraryProfiling = true;
    aeson.source = "1.5.0";
  };
}
```

?

PR #286, opened 2024-03-29
Reintroduce `buildFromSdist` (enable it by default)

### LINKED ISSUE #262: Enable `buildFromSdist` by default - to ensure release-worthiness
https://community.flake.parts/haskell-flake/settings#buildfromsdist

@roberth made a good point [here](https://github.com/srid/haskell-flake/pull/220#issuecomment-1918850608):

> haskell-flake has a role in CI to ensure release-worthiness. Building from sdist is an important part of that.

Currently, enabling `buildFromSdist` break things (see #253). But we should a) add tests for those (one exists for `jailbreak`, we need another one for `check`), b) find the root cause and fix the breakage, and c) re-enable it as the default.

srid:
Resolves #262


srid:
I tried using `mkOrder` here, but that didn't work. I want the whole settings code to be refactored, potentially as part of #285.

PR #291, opened 2024-04-02
fix: apply 'patches' after 'buildFromSdist'

### LINKED ISSUE #290: Patches are applied twice
The `haskell-flake` is trying to apply patches twice. The issue seems to be in the new `buildFromSDist` implementation. Setting `buildFromSDist` to `false` resolves the problem. For a reproduction, visit https://github.com/johnhampton/haskell-flakes-double-patch.

```nix
{
  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
    flake-parts.url = "github:hercules-ci/flake-parts";
    haskell-flake.url = "github:srid/haskell-flake";
  };
  outputs = inputs@{ self, nixpkgs, flake-parts, ... }:
    flake-parts.lib.mkFlake { inherit inputs; } {
      systems = nixpkgs.lib.systems.flakeExposed;

      imports = [ inputs.haskell-flake.flakeModule ];
      perSystem = { self', pkgs, config, ... }:
        {
          haskellProjects.default = {
            settings = {
              logict = {
                # buildFromSdist = false;
                patches = [ ./nix/patches/logict.patch ];
              };
            };
          };
        };
    };
}
```

```
error: builder for '/nix/store/2mbwfyxq1jvqiz3cq52dn8r9za3fia06-logict-0.8.1.0.drv' failed with exit code 1;
       last 10 log lines:
       > source root is logict-0.8.1.0
       > setting SOURCE_DATE_EPOCH to timestamp 1711982133 of file logict-0.8.1.0/test/Test.hs
       > warning: file logict-0.8.1.0/test/Test.hs may be generated; SOURCE_DATE_EPOCH may be non-deterministic
       > Running phase: patchPhase
       > applying patch /nix/store/r335m619yhwcqr0flp4ad9s6kihqqz3w-logict.patch
       > patching file logict.cabal
       > Reversed (or previously applied) patch detected!  Assume -R? [n]
       > Apply anyway? [n]
       > Skipping patch.
       > 1 out of 1 hunk ignored -- saving rejects to file logict.cabal.rej
       For full logs, run 'nix log /nix/store/2mbwfyxq1jvqiz3cq52dn8r9za3fia06-logict-0.8.1.0.drv'.
error: 1 dependencies of derivation '/nix/store/bp67yaxnr421c3whglrdgvy4a7dy03hx-haskell-flakes-double-patch-0.1.0.0.drv' failed to build
```

srid:
Resolves #290

### DIFF
--- a/nix/modules/project/settings/default.nix
+++ b/nix/modules/project/settings/default.nix
@@ -77,8 +77,8 @@ in
             # In future, we can refactor this as part of https://github.com/srid/haskell-flake/issues/285
             # NOTE: removeReferencesTo must apply *before* buildFromSdist, because the
             # later appears it fuck up the former otherwise.
-            impl = lib.attrsets.removeAttrs cfg.impl [ "buildFromSdist" "removeReferencesTo" ];
-            fns = lib.attrValues impl ++ [ cfg.impl.buildFromSdist cfg.impl.removeReferencesTo ];
+            impl = lib.attrsets.removeAttrs cfg.impl [ "buildFromSdist" "removeReferencesTo" "patches" ];
+            fns = lib.attrValues impl ++ [ cfg.impl.buildFromSdist cfg.impl.removeReferencesTo cfg.impl.patches ];
           in
           lib.pipe super.${name} (
             # TODO: Do we care about the *order* of overrides?


roberth:
That's a bit surprising. I'd only use `buildFromSdist` on local packages, and then it doesn't make much sense to add patches.
I suppose it could happen when importing a haskell-flake module from a dependency?

I don't have an opinion on whether patches should be applied before or after sdist. When applied before, the patches can be used to fix the creation of the sdist, whereas if you apply them after the sdist, you can fix the effects of a broken sdist.
I suppose the latter matches what you would do with a package from hackage, so I guess that might be a reason to prefer that.

roberth (approved):

srid:
> I'd only use `buildFromSdist` on local packages

Actually, that makes more sense.

Issue #293, opened 2024-04-02
Problems with `buildFromSdist`


srid:
`buildFromSdist` provides the benefit of checking release-worthiness (something not all Haskell projects care about, because they don't all care to release to Hackage).

But it does so at the expense of causing much pain:

- #238
- #283
- #292

Can we replace it with something that just works?

PR #294, opened 2024-04-02
fix: allow buildFromSdist to ignore Rust drvs

### LINKED ISSUE #292: `buildFromSdist` fails with Rust packages
The haskell package set can contain non-Haskell packages, like Rust libraries. These are currently set using `custom` option (but see #285) for better way, which is applied as the last option, but before `buildFromSdist`. Thus, `buildFromSdist` (because it is enabled by default since #286) will evaluate on these non-Haskell packages as well.

```
$ nix build github:shivaraj-bh/haskell-rust-ffi-template --override-input haskell-flake github:srid/haskell-flake/51bcbc27194e734a1524ac3d36e1539fba9a3baa
[..]
        at /nix/store/fjvkr94j3lwca8y66zccn15rjwzqn0rd-source/pkgs/development/haskell-modules/lib/compose.nix:308:25:

          307|    */
          308|   buildFromSdist = pkg: overrideCabal (drv: {
             |                         ^
          309|     src = "${sdistTarball pkg}/${pkg.pname}-${pkg.version}.tar.gz";

       … while calling 'overrideCabal'

         at /nix/store/fjvkr94j3lwca8y66zccn15rjwzqn0rd-source/pkgs/development/haskell-modules/lib/compose.nix:38:22:

           37|    */
           38|   overrideCabal = f: drv: (drv.override (args: args // {
             |                      ^
           39|     mkDerivation = drv: (args.mkDerivation drv).override f;

       error: attribute 'override' missing

       at /nix/store/fjvkr94j3lwca8y66zccn15rjwzqn0rd-source/pkgs/development/haskell-modules/lib/compose.nix:38:28:

           37|    */
           38|   overrideCabal = f: drv: (drv.override (args: args // {
             |                            ^
           39|     mkDerivation = drv: (args.mkDerivation drv).override f;
```

srid:
Resolves #292 as a hack, but what is the proper fix here for notorious (#293) `buildFromSdist`?

### DIFF
--- a/nix/modules/project/settings/all.nix
+++ b/nix/modules/project/settings/all.nix
@@ -318,11 +318,18 @@ in
         for release-worthiness.
       '';
       impl = enable:
+        let
+          # A version that doesn't break against Rust derivations
+          # https://github.com/srid/haskell-flake/issues/292
+          buildFromSdist' = drv:
+            if !lib.hasAttr "override" drv then drv else buildFromSdist drv;
+        in
         if enable then
-          (pkg: lib.pipe pkg [
-            buildFromSdist
-            (x: log.traceDebug "${name}.buildFromSdist ${x.outPath}" x)
-          ]) else null;
+          (pkg:
+            lib.pipe pkg [
+              buildFromSdist'
+              (x: log.traceDebug "${name}.buildFromSdist ${x.outPath}" x)
+            ]) else null;
     };
 
     removeReferencesTo = {


srid:
Closing in favour of #298

roberth:
> ignore Rust drvs

I thought I could ignore this but it freaks me out, somehow.

Is it really necessary to do this?
If there's a good reason to do it this way, it *might* be acceptable to insert them into the overlay, as a hacky implementation, but in that case please make it a real feature with a separate option, so that it doesn't break by itself, and so that it can be deprecated when a better solution comes along.

srid:
@roberth This PR got closed without being merged. Not sure what you are referring to. When you said *"Is it really necessary to do this?"* what you are referring to?

srid:
@roberth I think you are referring to the regression which is now fixed in #306

roberth:
> When you said _"Is it really necessary to do this?"_ what are you referring to?

Having rust files in the overlay, supporting that.

I apologize if I'm replying to the wrong thread or necrobumping unnecessarily.

srid:
> > When you said _"Is it really necessary to do this?"_ what are you referring to?
> 
> Having rust files in the overlay, supporting that.

This should be already supported through the `custom` option; see

https://github.com/shivaraj-bh/haskell-rust-ffi-template/blob/0ea595271103161184332ad4c5322fbc1c93a711/haskell/default.nix#L7

Do you have a better interface in mind?

(There's a proposal that incidentally gets rid of this `custom` option; see #285)

PR #298, opened 2024-04-03
buildFromSdist: enable only for local packages

### LINKED ISSUE #292: `buildFromSdist` fails with Rust packages

srid:
https://github.com/srid/haskell-flake/pull/291#issuecomment-2033350702

Resolves #292

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Breaking changes
   - #221: Switch from `buildFromCabalSdist` to `buildFromSdist`, to allow using non-standard package sets (wherein `cabal-install` is otherwise built without using user's overrides)
-    - #253: Enable controlling `buildFromSdist` through `settings.<name>.buildFromSdist`. (This was turned off by default originally, but was turned on by default in #286)
+    - #253: Enable controlling `buildFromSdist` through `settings.<name>.buildFromSdist`. (This was turned off by default originally, but was turned on by default in #286, limited to local packages in #298)
 - Enhancements
   - `settings` module:
     - #210: Add `extraLibraries` to `settings` module.
--- a/nix/modules/project/defaults.nix
+++ b/nix/modules/project/defaults.nix
@@ -84,21 +84,18 @@ in
       '';
       default =
         let
-          globalSettings = {
-            # Make sure all files we use are included in the sdist, as a check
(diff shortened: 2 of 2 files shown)


shivaraj-bh:
was it intended to be `defaults.settings.defined` here? or was it supposed to be `defaults.settings.local`?

srid:
It should be `defined`.

Wait, the last PR reverted this change, no?

I see this under `settings.all`:

https://github.com/srid/haskell-flake/blob/e43bc631b413078ffe7b772850005596f7295050/nix/modules/project/defaults.nix#L197

shivaraj-bh:
> Wait, the last PR reverted this change, no?

Shoot, I messed up while resolving merge conflict

PR #382, opened 2024-12-08
feat: `cabal2nix` integration for IFD-less evalutation

### LINKED ISSUE #144: IFD-less evaluation of local packages

rsrohitsingh682:
Addresses #144 just for local packages. IFD will still be used for Hackage packages (`callHackage`) and custom source inputs. 

Example: https://github.com/srid/haskell-template/pull/146


nix/build-haskell-package.nix
```diff
@@ -22,6 +22,20 @@ let
       name = "${name}";
       inherit src;
     };
+
+  callPackageKeepDeriver = src: args:
```
srid, line 26: You want to add a comment explaining why this function is here.

It appears to be a copy of:

https://github.com/NixOS/nixpkgs/blob/d30e9cf3be1cc72f99d5f67772ce961b84126adf/pkgs/development/haskell-modules/make-package-set.nix#L170

Is there no way of directly using it from nixpkgs?

nix/build-haskell-package.nix
```diff
@@ -31,6 +45,19 @@ lib.pipe root
...
     (x: log.traceDebug "${name}.mkNewStorePath ${x.outPath}" x)
 
-    (root: self.callCabal2nix name root { })
-    (x: log.traceDebug "${name}.cabal2nixDeriver ${x.cabal2nixDeriver.outPath}" x)
+     (root:
+        # Check if cached cabal2nix generated nix expression is present,
+        # if present use it with callPackage
+        # to avoid IFD
+        let pkg = 
+          if lib.pathExists (lib.concatStringsSep "/" [ root "default.nix" ])
```
srid, line 53: The name of this file should be configurable by the user, much like rust-flake's [`crateNixFile`](https://github.com/juspay/rust-flake/blob/cdcm17a1d009674a9a23c7e4749baf474d2419f04/nix/modules/flake-module.nix#L54).

Also, the default value of this option should not be `default.nix` (people use it for a lot of other things), but rather something ore unique, like `cabal.nix`.

nix/build-haskell-package.nix
rsrohitsingh682: Will check that.

nix/build-haskell-package.nix
srid: You can open a PR on https://github.com/cachix/git-hooks.nix adding an option that allows the user to override it.

rsrohitsingh682: https://github.com/cachix/git-hooks.nix/pull/531

nix/build-haskell-package.nix
```diff
@@ -22,15 +22,37 @@ let
...
+      else
+        let
+          pkg =
+            (log.traceDebug "${name}.callCabal2nix ${root}")
+              self.callCabal2nix
+              name
+              root
+              { };
+        in
+        (log.traceDebug "${name}.cabal2nixDeriver ${pkg.cabal2nixDeriver.outPath}" pkg)
```
srid, line 56: I think should only be logged when `callCabal2nix` is being used.

rsrohitsingh682: Yeah. This is exactly that happens. 
` (log.traceDebug "${name}.cabal2nixDeriver ${pkg.cabal2nixDeriver.outPath}" pkg)` gets logged only when we use `callCabal2nix`.

Logs from `nammayatri`
https://gist.github.com/rsrohitsingh682/243463dbffaeea5fc87f488a997736d1

nix/build-haskell-package.nix
```diff
@@ -22,15 +22,37 @@ let
...
```
srid, line 56:
```suggestion
        lib.pipe (self.callPackage path {})
          [ (pkg: log.traceDebug "${name}.callCabal2nix root=${root} deriver=${pkg.cabal2nixDeriver.outPath}")
          ]
```

srid (commented):
LGTM. 

As a last thing, could [add a test](https://github.com/srid/haskell-flake/tree/master/test)?

srid (approved):
LGTM. Just wanting on the other PRs to be merged. And then I'll test it on some of my projects before merging.

PR #384, opened 2024-12-17
Add `stan` setting to output static analysis report

shivaraj-bh:
image

Generating this report is just:
```nix
# Inside haskellProjects.<project-name>
{
  settings.<package-name>.stan = true;
}
```

### DIFF
--- a/doc/settings.md
+++ b/doc/settings.md
@@ -70,3 +70,13 @@ Newer versions of [nixpkgs] provide `buildFromSdist` to build your package from 
 > If you encounter issues with `buildFromSdist` you can disable it by setting `settings.<name>.buildFromSdist` to `true`.
 
 [nixpkgs]: https://nixos.asia/en/nixpkgs
+
+### `stan`
+
+Run **ST**atic **AN**alysis on the package using [stan] and generate an HTML report. The report is created in the `/nix/store` path alongside your package outputs.
+
+> [!note] stan configuration  
+> This setting looks for a `.stan.toml` file in the root of the package source. See a sample [.stan.toml] configuration for reference.
+
+[stan]: https://github.com/kowainik/stan
+[.stan.toml]: https://github.com/kowainik/stan/blob/main/.stan.toml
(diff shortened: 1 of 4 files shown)


shivaraj-bh:
From the logs of running the test above:

image

And the HTML report is in the `$out`:
image

doc/settings.md
```diff
@@ -70,3 +70,13 @@ Newer versions of [nixpkgs] provide `buildFromSdist` to build your package from
 > If you encounter issues with `buildFromSdist` you can disable it by setting `settings.<name>.buildFromSdist` to `true`.
 
 [nixpkgs]: https://nixos.asia/en/nixpkgs
+
+### `stan`
+
+Run **ST**atic **AN**alysis on the package using [stan] and generate an HTML report. The report is created in the `/nix/store` path alongside your package outputs.
+
+> [!note] stan configuration  
+> This setting looks for a `.stan.toml` file in the root of the package source. See a sample [.stan.toml] configuration for reference.
```
shivaraj-bh, line 79: A bit of hardcoding here. I guess we could eliminate this by making the stan setting a submodule (providing an option to configure custom configuration path) instead of just a boolean, but I doubt that’s necessary.

srid:
@shivaraj-bh Do we also use the HLS integration?

https://github.com/haskell/haskell-language-server/pull/2908

srid:
Trying this on vira project I hit https://github.com/kowainik/stan/issues/392 ... I wonder if this happens on internal projects too.

doc/settings.md
srid: I followed these instructions and added `.stan.toml` to the root:

https://github.com/juspay/vira/commit/53a42bf8a6bf627e9690ea402ce67d2036b59295

However, it doesn't seem to be used, because the generated report still doesn't exclude the StrictData suggestion:

image

It'd be good to demonstrate this in a sample repo, like https://github.com/srid/haskell-template - along with `.stan.toml`.

srid:
Another thing worth considering: add a flake check that will fail if project health is not 100%. Poeple can always add excludes to their stan config if there's a good reason to reject certain suggestions.

doc/settings.md
srid: cf. https://github.com/srid/haskell-template/pull/147

shivaraj-bh:
> Do we also use the HLS integration?

No

PR #411, opened 2025-03-28
docs(gotchas): Overriding `libssh2`

shivaraj-bh:


doc/gotchas.md
```diff
@@ -0,0 +1,27 @@
...
+```nix
+# In `haskellProjects.default`
+{
+  settings = {
+    libssh2 = {
+      broken = false;
+      # Overriding the package with `packages.libssh2.source = "0.2.0.9"` results in infinite recursion.
+
+      # Possibly having to do with `cabal2nix` not understading that [`libssh2` in `pkgconfig-depends` of `libssh2.cabal`](https://github.com/portnov/libssh2-hs/blob/bf7cbe643c7f4fb4fad3963705feb8351471eb01/libssh2/libssh2.cabal#L70)
+      # is not self-referential.
```
srid, line 15: Can you move this above as Markdown paragraph below the heading? Without reading the code, the 'gotcha' section is stripped of any explanation as to its existance.

shivaraj-bh: @srid moved in https://github.com/srid/haskell-flake/pull/411/commits/a1c12d1bbbe3b79e2caf78229c7894ffe2158d97

PR #418, opened 2025-04-20
Add `extraCabal2nixOptions` and `cabalFlags` to `packages.<package>`

np:
Settings such as `settings.<package>.cabalFlags` happen after `cabal2nix` which can be too late when the dependencies depend on a flag.

The option `packages.<package>.cabalFlags` can safely be used instead of `settings.<package>.cabalFlags` in most situations.

Additionally the option `packages.<package>.extraCabal2nixOptions` provides a direct access the options given to cabal2nix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Enhancements
   - #382: Support for cabal2nix generated expressions (avoiding IFD for local packages)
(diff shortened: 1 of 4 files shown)


srid (commented):
Hey @np - thanks for the PR, this is quite an useful feature.

CI is failing with the following. Note that you can run `nix run nixpkgs#omnix ci run` locally to run CI. If that doesn't work, run `nix --accept-flake-config run github:juspay/omnix ci run`

image

PR #420, opened 2025-04-23
buildFromSdist: Test if user can override the `jailbreak` setting

shivaraj-bh:
Adds one of the tests mentioned in the description of https://github.com/srid/haskell-flake/issues/262

The existing jailbreak test `foo.jailbreak = true;` doesn't test the expected behaviour because `foo`'s `cabal` file has no dependencies requiring jailbreak

### DIFF
--- a/test/simple/flake.nix
+++ b/test/simple/flake.nix
@@ -51,8 +51,12 @@
               jailbreak = true;
               cabalFlags.blah = true;
             };
-            # Test STatic ANalysis report generation
-            haskell-flake-test.stan = true;
+            haskell-flake-test = {
+              # Test STatic ANalysis report generation
+              stan = true;
+              # Test if user's setting overrides the `jailbreak = false;` override by `buildFromSdist`
+              jailbreak = true;
+            };
           };
           devShell = {
             tools = hp: {
--- a/test/simple/flake.nix
+++ b/test/simple/flake.nix
@@ -54,7 +54,9 @@
             haskell-flake-test = {
               # Test STatic ANalysis report generation
               stan = true;
-              # Test if user's setting overrides the `jailbreak = false;` override by `buildFromSdist`
+              # Test if user's setting overrides the `jailbreak = false;` override by `buildFromSdist`.
+              # 
+              # This jailbreak ignores the unsatisfiable version constraints on the library `foo`.
               jailbreak = true;
             };
           };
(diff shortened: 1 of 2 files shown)


test/simple/haskell-flake-test.cabal
```diff
@@ -14,6 +14,6 @@ executable haskell-flake-test
     main-is:          Main.hs
     build-depends:    
         base,
-        foo
+        foo >= 0.2  -- To test `jailbreak` in flake
```
srid, line 17: Added a comment. I assume that's correct.

srid: @shivaraj-bh Could you confirm or update the comment?

test/simple/flake.nix
```diff
@@ -51,8 +51,12 @@
               jailbreak = true;
               cabalFlags.blah = true;
             };
-            # Test STatic ANalysis report generation
-            haskell-flake-test.stan = true;
+            haskell-flake-test = {
+              # Test STatic ANalysis report generation
+              stan = true;
+              # Test if user's setting overrides the `jailbreak = false;` override by `buildFromSdist`
+              jailbreak = true;
```
srid, line 58: Add a comment here saying what exactly is being jailbroken.

test/simple/haskell-flake-test.cabal
```diff
@@ -14,6 +14,6 @@ executable haskell-flake-test
     main-is:          Main.hs
     build-depends:    
         base,
-        foo
+        foo >= 0.2
```
srid, line 17: Add a comment here saying why `>= 0.2` exists and, especially, how it relates to the jailbreak flag in the flake.

srid (changes requested):

test/simple/flake.nix
shivaraj-bh: https://github.com/srid/haskell-flake/pull/420/commits/1a4f4f6ea1ac6c87c238cbfd72cf8d2358b5a3ec

test/simple/haskell-flake-test.cabal
shivaraj-bh: https://github.com/srid/haskell-flake/pull/420/commits/1a4f4f6ea1ac6c87c238cbfd72cf8d2358b5a3ec

PR #458, opened 2025-09-15
Apply the custom overlay first

### LINKED ISSUE #438: make `custom` the first overlay?
I think changing from "code 1" to "code 2" improves UX.

```nix
# code 1
settings.foo = { super, ... }: {
  custom = _: pkgs.haskell.lib.compose.markUnbroken super.foo_0_7_0_0;
};

# code 2
settings.foo = { super, ... }: {
  broken = false;
  custom = _: super.foo_0_7_0_0;
};
```

Changing the order of `custom` is also [mentioned][1] in the comments.  I am willing to create a PR for this if this change is desirable.

[1]: https://github.com/srid/haskell-flake/blob/39065472d2587af93a502423276bfb98c2c6fb09/nix/modules/project/settings/default.nix#L85-L93

jian-lin:
Close #438

### DIFF
--- a/nix/modules/project/settings/default.nix
+++ b/nix/modules/project/settings/default.nix
@@ -75,6 +75,7 @@ in
               });
             }).config;
             cfg = traceSettings name cfg';
+            # NOTE: apply custom first for better UX
             # HACK: buildFromSdist must apply *last*
             # cf. https://github.com/srid/haskell-flake/pull/252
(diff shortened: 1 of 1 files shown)


jian-lin:
For the `markUnbroken` example in #438, making it first works.  Making it last would discard all other overlays.  Not sure if there are cases when only making it last works though.

PR #495, opened 2026-03-28
Support using haskell-flake without flake-parts

srid:
**haskell-flake can now be used standalone**, without requiring flake-parts as a framework. A new `lib.evalHaskellProject` function evaluates project configuration directly via `lib.evalModules`, returning the same outputs (packages, devShell, overlay, apps, checks) that the flake-parts module produces.

The project submodule was already *nearly* self-contained — it only depended on flake-parts through the `self` specialArg used as default for `projectRoot`. This PR removes that coupling: `projectRoot` no longer has a default in the submodule, and the flake-parts wrapper injects `projectRoot = lib.mkDefault self` to preserve existing behavior. *This also fixes a latent bug in `hls-check.nix` where `self` was used instead of `config.projectRoot`, which would be wrong whenever `projectRoot` is overridden.*

Follows the pattern established by [process-compose-flake#80](https://github.com/Platonic-Systems/process-compose-flake/pull/80).

### Try it locally

```nix
let
  project = (haskell-flake.lib { inherit pkgs; }).evalHaskellProject {
    projectRoot = self;
  };
in project.packages.mypackage.package
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Enhancements
+  - #495: Add `lib.evalHaskellProject` for using haskell-flake without flake-parts. Also fixes `hls-check.nix` to use `config.projectRoot` instead of `self`.
   - #382: Support for cabal2nix generated expressions (avoiding IFD for local packages)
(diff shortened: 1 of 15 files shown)


srid:
Tested in https://github.com/srid/hint-demo/commit/6a3365859da78c171098bccae887abb47f95e414

